PR #1, opened 2021-03-29
Raccoon proto

NNcrawler:
Hi, this MR includes proto files used by Raccoon.
I also added [buf](https://buf.build/) for proto-related operations.


odpf/raccoon/Event.proto
```diff
@@ -0,0 +1,31 @@
...
+
+option java_multiple_files = true;
+option java_package = "io.odpf.raccoon";
+option java_outer_classname = "EventProto";
+
+/*
+ `Event` defines wrapper of the event you want to push to Raccoon
+
+ The event populated here is in form of bytes. It means you can use any data format or schema
+ you want for your event. It can be as simple as JSON encoded key-value pair or complex
```
chakravarthyvp, line 13: This sentence could be removed mas.

chakravarthyvp, line 13:
```suggestion
technically it's right. But let's keep it for use with protobuf mas as the intention is to move away from json for anybody using it.
```

PR #2, opened 2021-03-31
Feature: Add publishing to github packages and Optimus metadata proto

rootcss:


src/odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,77 @@
...
+option java_outer_classname = "OptimusLog";
+option go_package = "github.com/odpf/proton/metadata/resource";
+
+import "google/protobuf/timestamp.proto";
+
+message OptimusLogKey {
+  string urn = 1;
+}
+
+message OptimusLogMessage {
```
kushsharma, line 15: Optimus will start managing warehouse resources as well, like tables/views, etc. How do you think they fit here? We should have a message for Job specifically.

src/odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,77 @@
...
+  google.protobuf.Timestamp end_date = 2;
+  string interval = 3;
+}
+
+message OptimusJobBehavior {
+  bool depends_on_past = 1;
+  bool catchup = 2;
+}
+
+message OptimusConfig {
```
kushsharma, line 74: Similar to what I have stated above, this message name if pretty generic.

src/odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,77 @@
...
+
+message OptimusLogKey {
+  string urn = 1;
+}
+
+message OptimusLogMessage {
+  string urn = 1;
+  int32 version = 2;
+  string description = 3;
+  repeated OptimusConfig labels = 4;
```
kushsharma, line 19: This is an incorrect representation of labels datatype although the message schema is the same.

src/odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,77 @@
...
+  OptimusJobSchedule schedule = 7;
+  OptimusJobBehavior behaviour = 8;
+
+  repeated OptimusHook hooks = 9;
+  repeated OptimusJobDependency dependencies = 10;
+
+  google.protobuf.Timestamp event_timestamp = 100;
+}
+
+message OptimusTask {
```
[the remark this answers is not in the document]
rootcss: I've renamed it to OptimusJobTask.

src/odpf/proton/metadata/resource/OptimusLog.proto
rootcss: renamed it to OptimusJobMetadata

src/odpf/proton/metadata/resource/OptimusLog.proto
rootcss: I've introduced following:
```proto
message OptimusJobLabel {
  string name = 1;
  string value = 2;
}

message OptimusJobTaskConfig {
  string name = 1;
  string value = 2;
}

message OptimusJobHookConfig {
  string name = 1;
  string value = 2;
}
```

src/odpf/proton/metadata/resource/OptimusLog.proto
rootcss: I've introduced following for it:
```proto
message OptimusJobLabel {
  string name = 1;
  string value = 2;
}
```

odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,87 @@
...
+option go_package = "github.com/odpf/proton/metadata/resource";
+
+import "google/protobuf/timestamp.proto";
+
+message OptimusJobMetadataKey {
+  string urn = 1;
+}
+
+message OptimusJobMetadata {
+  string urn = 1;
```
kushsharma, line 16: How exactly do you define `urn` here? Is this just the name? Or it's really unique across projects?

odpf/proton/metadata/resource/OptimusLog.proto
```diff
@@ -0,0 +1,87 @@
...
+package odpf.proton.metadata.resource;
+
+option java_package = "io.odpf.proton.metadata.resource";
+option java_outer_classname = "OptimusLog";
+option go_package = "github.com/odpf/proton/metadata/resource";
+
+import "google/protobuf/timestamp.proto";
+
+message OptimusJobMetadataKey {
+  string urn = 1;
```
kushsharma, line 12: Can you give an example what this will look like? How are we mapping it with `tenants`?

odpf/proton/metadata/resource/OptimusLog.proto
rootcss: as discussed we'll keep it like this, and it'll be unique across all tenants.
```
tenant_name::job/<name>
tenant_name::table/<name>
tenant_name::view/<name>
```

.github/build.yml
```diff
@@ -0,0 +1,17 @@
+name: Build
```
NNcrawler, line 1: This should be inside `.github/workflows/`. CMIIW

.github/package.yml
```diff
@@ -0,0 +1,32 @@
+name: Package
```
NNcrawler, line 1: This should be inside `.github/workflows/`. CMIIW

.github/build.yml
rootcss: correct. fixed it now.

.github/package.yml
rootcss: correct. fixed it now.

NNcrawler:
LGTM

ravisuhag:
@rootcss Can we use just protoc and github actions instead of gradle for compiling jar and publishing?

rootcss:
@ravisuhag yes, we can do that too. We'll have to add steps like:
- download protoc binary or use `apt` to install, since we're using ubuntu.
- compile protos using something like: `protoc -I=./odpf --java_out=./output/ $(find ./odpf -iname "*.proto")`
- since above generates java files, we'll have to compile those to create the Jar. (maven?)
- and then publish.

I think, Using gradle simplifies it overall and easier to test out locally since compilation doesn't sit inside the github workflows. gradle also allows us to keep any Java related things inside the project if needed in future. WDYT?

rootcss:
Can we merge this @ravisuhag @NNcrawler ? Need to use the package in Shenron and Optimus.

PR #9, opened 2021-05-17
feat: add client RPC attributes to optimus Runtime proto

rootcss:


odpf/optimus/RuntimeService.proto
```diff
@@ -142,6 +178,10 @@ message ProjectSpecification {
     map<string, string> config = 2;
 }
 
+message ClientSpecification {
```
kushsharma, line 181: Shouldn't we need client-specific configs here as well?

PR #11, opened 2021-06-07
feat: add optimus replay request

mauliksoneji:


odpf/optimus/runtime_service.proto
```diff
@@ -168,6 +168,11 @@ service RuntimeService {
             body: "*"
         };
     }
+    rpc ReplayDryRun(ReplayDryRunRequest) returns (ReplayDryRunResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/project/{project_name}/job/{job_name}/replayDryRun"
```
kushsharma, line 173: Can we use lower case letters with hyphen(-) separator(spinal-case) here?

odpf/optimus/runtime_service.proto
```diff
@@ -544,3 +549,22 @@ message UpdateResourceResponse {
...
+    string end_date = 5;
+}
+
+message ReplayDryRunResponse {
+    bool success = 1;
+    ReplayExecutionTreeNode response = 2;
+}
+
+message ReplayExecutionTreeNode {
+    string jobName = 1;
```
kushsharma, line 567: job_name

odpf/optimus/runtime_service.proto
mauliksoneji: updated

odpf/optimus/runtime_service.proto
mauliksoneji: updated

PR #17, opened 2021-07-12
feat: add guardian proto

rahmatrhd:


ravisuhag (changes requested):

odpf/guardian/guardian.proto
```diff
@@ -0,0 +1,370 @@
...
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/odpf/proton/guardian/go_package";
+option java_multiple_files = true;
+option java_package = "io.odpf.proton.guardian";
+option java_outer_classname = "ServiceManager";
+
+// These annotations are used when generating the OpenAPI file.
+option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
```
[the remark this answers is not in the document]
rahmatrhd: https://github.com/grpc-ecosystem/grpc-gateway/pull/2147 the support for v3 still under development

odpf/guardian/guardian.proto
```diff
@@ -0,0 +1,370 @@
...
+        };
+    }
+
+    rpc ListApprovals(ListApprovalsRequest) returns (ListApprovalsResponse) {
+        option (google.api.http) = {
+            get: "/appeals/approvals"
+        };
+    }
+
+    rpc MakeAction(MakeActionRequest) returns (MakeActionResponse) {
```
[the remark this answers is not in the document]
rahmatrhd: renamed to `UpdateApproval`

ravisuhag:
LGTM. @kushsharma Can you also cross-check for any conventions we are following in proton?

odpf/guardian/guardian.proto
```diff
@@ -0,0 +1,375 @@
...
+    Action action = 3;
+}
+
+message UpdateApprovalResponse {
+    Appeal appeal = 1;
+}
+
+// Provider contains information about external data provider such as BigQuery, Metabase, etc., credentials, policy, and allowed roles 
+message Provider {
+    uint32 id = 1;
```
kushsharma, line 253: I see we are using int ids everywhere, are these DB auto-increment ids? Why not use UUID?

odpf/guardian/guardian.proto
```diff
@@ -0,0 +1,375 @@
...
+    string type = 2;
+    string urn = 3;
+
+    message ProviderConfig {
+        string type = 1;
+        string urn = 2;
+        google.protobuf.Struct labels = 3;
```
kushsharma, line 260: I see labels as `key,value` pairs of string, any reason why we chose Struct for this? We should limit the usage of `Struct` to only places where we know for sure we can't use concrete types I think as it adds inconvenience while parsing it.

kushsharma (changes requested):

kushsharma:
@ravisuhag are we not versioning our REST APIs? Not versioning protobufs is fine I think as they get imported and get versioned while importing but REST APIs have no convention of knowing if something has breaking changes right?

odpf/guardian/guardian.proto
rahmatrhd: Yes DB auto-increment, no specific reason why we're not using UUID, just following traditional ways for creating entity. What would make it better to use UUID?

odpf/guardian/guardian.proto
rahmatrhd: The type we have for labels in guardian right now is `map[string]interface{}`, using Struct to fulfill that.

ravisuhag: I think this is better moved to [string][string] labels should not be In any other type.

ravisuhag:
+1 to versioning. Shall we right away start with v1 as folder as well and move protos to that? or any other better idea?

odpf/guardian/guardian.proto
rahmatrhd: changing the proto type for labels to `map<string, string>`. Will change the labels type in guardian to `map[string]string` as well

ravisuhag:
let's skip `/api ` if needed service should handle it.

odpf/guardian/guardian.proto
```diff
@@ -0,0 +1,375 @@
...
+        version: "0.1";
+    };
+    external_docs: {
+        description: "Guardian server";
+    }
+    schemes: HTTP;
+};
+
+// WARNING: This is still in active development and can have breaking changes
+service Guardian {
```
kushsharma, line 26: It's a general guideline to add `Service` suffix to services to distinguish them from proto messages.

odpf/guardian/guardian.proto
kushsharma: - can't enumerate values
- can't infer how many values are present in each table
- don't have to talk to the database to construct a value, useful in certain ORM situations
- perfectly indexable by most dbs
- easier to shard when we need to distribute data into multiple dbs
- unique across multiple tables/databases/space-time continuum
- costly on size but storage is cheap

kushsharma:
Yeah, we can version protobufs at directory/package level as well although once we have stable services, function definition hardly changes, only message fields do and that can be handled by only introducing new fields with new indices and supporting old fields as well. But yeah I am onboard with directory level versioning as well.
I need to refactor `optimus` protobufs though, went with the simplest one when started and now it has so many things going on.
Like this
```
.
└── proto
    └── foo
        └── bar
            ├── bat
            │   └── v1
            │       └── bat.proto // package foo.bar.bat.v1
            └── baz
                └── v1
                    ├── baz.proto         // package foo.bar.baz.v1
                    └── baz_service.proto // package foo.bar.baz.v1
```

ravisuhag:
@rahmatrhd lets make the discussed changes and then we can merge.

ravisuhag:
@rahmatrhd Are we moving protos to V1 folder?

kushsharma:
> @rahmatrhd @kushsharma Are we moving protos to V1 folder?

yeah we should

rohilsurana:
`package odpf.guardian;` => `package odpf.guardian.v1;`
`option go_package = "github.com/odpf/proton/guardian";` => `option go_package = "github.com/odpf/proton/guardian/v1;guardianv1";`

same goes for the directory as well.

rohilsurana:
Let's follow semver for `openapiv2_swagger.info.version`.

PR #19, opened 2021-07-20
feat: add metadata models

StewartJingga:


odpf/meta/common/Resource.proto
```diff
@@ -0,0 +1,14 @@
...
+package odpf.meta.common;
+
+option java_package = "io.odpf.proton.meta.common";
+option java_outer_classname = "Resource";
+option go_package = "github.com/odpf/proton/meta/common";
+
+message Resource {
+  string urn = 1;
+  string name = 2;
+  string source = 3;
```
kushsharma, line 12: can you give an example of what this source will contain?

odpf/meta/facets/Ownership.proto
```diff
@@ -0,0 +1,17 @@
...
+option java_package = "io.odpf.proton.meta.facets";
+option java_outer_classname = "Ownership";
+option go_package = "github.com/odpf/proton/meta/facets";
+
+message Ownership {
+  repeated Owner owners = 7;
+}
+
+message Owner {
+  string urn = 1;
```
kushsharma, line 14: What could be `urn` for owners?

odpf/meta/facets/Columns.proto
```diff
@@ -0,0 +1,19 @@
...
+option go_package = "github.com/odpf/proton/meta/facets";
+
+message Columns {
+  repeated Column columns = 1;
+}
+
+message Column {
+  string name = 1;
+  string description = 2;
+  string data_type = 3;
```
kushsharma, line 16: why not just `type`?

odpf/meta/facets/Columns.proto
```diff
@@ -0,0 +1,19 @@
...
+  bool is_nullable = 4;
+  int64 length = 5;
```
kushsharma, line 18: what are we storing in length?

odpf/meta/facets/Custom.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package odpf.meta.facets;
+
+option java_package = "io.odpf.proton.meta.facets";
+option java_outer_classname = "Custom";
+option go_package = "github.com/odpf/proton/meta/facets";
+
+message Custom {
+  map<string,string> custom_properties = 1;
```
kushsharma, line 10: just call it `properties`
Also I can imagine this could contain `key -> []string` use cases. I think its better if we use `Value` here

odpf/meta/User.proto
```diff
@@ -0,0 +1,48 @@
...
+import "odpf/meta/common/Timestamp.proto";
+import "odpf/meta/common/Event.proto";
+
+message UserKey {
+  string urn = 1;
+}
+
+message User {
+  string urn = 1;
+  string source = 2;
```
kushsharma, line 20: whats this source used for?

odpf/meta/facets/Ownership.proto
ravisuhag: User UUID

odpf/meta/common/Resource.proto
ravisuhag: BigQuery, Kafka etc.

odpf/meta/User.proto
ravisuhag: Google, Shield, Okta etc.

odpf/meta/facets/Columns.proto
StewartJingga: `Column.type` might mean something else while `Column.data_type` clearly talking about the data type as in`varchar` or `int`.
wdyt?

odpf/meta/facets/Columns.proto
StewartJingga: for `varchar` data type can be something like `255`.

odpf/meta/Chart.proto
```diff
@@ -0,0 +1,36 @@
...
+}
+
+message Chart {
+  string urn = 1;
+  string name = 2;
+  string type = 3;
+  string source = 4;
+  string description = 5;
+  string url = 6;
+  string raw_query = 7;
```
irainia, line 27: For Grafana, there is a possibility of more than one query. Need to consider that as well. Suggestion, we can use array of string.

irainia: Also there is a need to know for which data source the query is run against. For example, in Grafana, a panel can have multiple queries, but the data source should only be one.

StewartJingga: > Also there is a need to know for which data source the query is run against. For example, in Grafana, a panel can have multiple queries, but the data source should only be one.

added `string data_source` on chart to solve this.

> For Grafana, there is a possibility of more than one query. Need to consider that as well. Suggestion, we can use array of string.

yeah agree, we need to collect more dashboard sample first though to make sure the changes would be generic.
for example, atm metabase card only has one query.

for now I would suggest adding extra queries to custom facets e.g. `raw_query_2`. Once we see a lot of chart has multiple queries then we can remodel it.

But I would assume most time-series dashboard would have multiple queries. 🤔

odpf/meta/facets/Columns.proto
kushsharma: Sounds good 👍  but then we should follow this in `length` field as well right?

odpf/meta/facets/Tags.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package odpf.meta.facets;
+
+option java_package = "io.odpf.proton.meta.facets";
+option java_outer_classname = "Tags";
+option go_package = "github.com/odpf/proton/meta/facets";
+
+message Tags {
```
kushsharma, line 9: Do we really need a facet for `Tag`? Do you see in the near future anyone adding any more fields other than this map?

StewartJingga: hmm good question 🤔 , as of now I havent found our use cases yet.
The tag facet is inspired from evaluating how Datahub and Amundsen modelling their metadata.

Im fine right now if we want to remove tag facet. wdyt @ravisuhag ?

odpf/meta/facets/Columns.proto
StewartJingga: imo column length is pretty common but maybe it's just me and if you didnt catch it the first time probably it needs to be renamed.

what would be clearer, `field_length`, `data_length`?

odpf/meta/facets/Tags.proto
ravisuhag: Yeah. should be fine.

StewartJingga (commented):
the new folder structure is `odpf/entities/[common,facets,resources]`.
do you think we should somehow mention metadata somewhere?

`odpf/entities` might mean all entities/models for odpf instead of metadata.

ravisuhag:
> the new folder structure is `odpf/entities/[common,facets,resources]`.
> do you think we should somehow mention metadata somewhere?
> 
> `odpf/entities` might mean all entities/models for odpf instead of metadata.

Ideally, I was thinking of actually having all entities of odpf but only from a metadata point of view. Let me think if I can add that context somehow.

PR #21, opened 2021-07-27
feat: add replay status schema in optimus runtime_service proto

arinda-arif:


odpf/optimus/runtime_service.proto
```diff
@@ -187,6 +187,11 @@ service RuntimeService {
             post: "/v1/project/{project_name}/job/{job_name}/replay"
```
kushsharma, line 187: Can you also check if this endpoint is working as expected? Because I don't see the `body` field being used here.

odpf/optimus/runtime_service.proto
```diff
@@ -187,6 +187,11 @@ service RuntimeService {
...
         };
     }
+    rpc GetReplayStatus(ReplayStatusRequest) returns (ReplayStatusResponse) {
```
kushsharma, line 190: according to proto guides, each message in rpc should be prefixed by its method name so this should be 
```
GetReplayStatusRequest/GetReplayStatusResponse
```

odpf/optimus/runtime_service.proto
```diff
@@ -622,6 +627,30 @@ message ReplayResponse {
...
+message ReplayStatusRun {
+    google.protobuf.Timestamp run = 1;
+    string state = 2;
+}
+
+message ReplayStatusRequest {
+    string id = 1;
+    string job_name = 2;
+    string project_name = 3;
+    string namespace = 4;
```
kushsharma, line 651: why do we need namespace?

odpf/optimus/runtime_service.proto
arinda-arif: have checked this and it wasn't working, so have added body. did the same for replay dry run too

odpf/optimus/runtime_service.proto
arinda-arif: updated

odpf/optimus/runtime_service.proto
arinda-arif: updated

PR #25, opened 2021-08-23
feat: add optimus replay list

arinda-arif:
Add Optimus get replay list rpc & message models.

### DIFF
--- a/odpf/optimus/runtime_service.proto
+++ b/odpf/optimus/runtime_service.proto
@@ -196,7 +196,7 @@ service RuntimeService {
     }
     rpc GetReplayList(GetReplayListRequest) returns (GetReplayListResponse) {
         option (google.api.http) = {
-            get: "/v1/project/{project_name}/replay/{id}"
+            get: "/v1/project/{project_name}/replay/list"
         };
     }
     // TODO(kush.sharma): disabled ATM
--- a/odpf/optimus/runtime_service.proto
+++ b/odpf/optimus/runtime_service.proto
@@ -194,6 +194,11 @@ service RuntimeService {
             get: "/v1/project/{project_name}/job/{job_name}/replay/{id}"
         };
     }
+    rpc GetReplayList(GetReplayListRequest) returns (GetReplayListResponse) {
+        option (google.api.http) = {
+            get: "/v1/project/{project_name}/replay/{id}"
+        };
+    }
     // TODO(kush.sharma): disabled ATM
     //rpc DeleteResource(DeleteResourceRequest) returns (DeleteResourceResponse) {}
 
@@ -661,3 +666,20 @@ message RegisterJobEventRequest {
 }
 
 message RegisterJobEventResponse {}
+
+message GetReplayListRequest {
+    string project_name = 1;
+}
+
+message GetReplayListResponse {
+    repeated ReplaySpec replay_list = 1;
+}
+
+message ReplaySpec {
+    string id = 1;
+    string job_name = 2;
+    string start_date = 3;
+    string end_date = 4;
+    string state = 5;
+    google.protobuf.Timestamp created_at = 6;
+}
--- a/odpf/optimus/runtime_service.proto
+++ b/odpf/optimus/runtime_service.proto
@@ -179,7 +179,7 @@ service RuntimeService {
     }
     rpc ReplayDryRun(ReplayRequest) returns (ReplayDryRunResponse) {
         option (google.api.http) = {
-            post: "/v1/project/{project_name}/replay/{job_name}/dryrun"
+            post: "/v1/project/{project_name}/replay/{job_name}:dryrun"
             body: "*"
         };
     }
--- a/odpf/optimus/runtime_service.proto
+++ b/odpf/optimus/runtime_service.proto
@@ -177,7 +177,7 @@ service RuntimeService {
             body: "*"
         };
     }
-    rpc ReplayDryRun(ReplayRequest) returns (ReplayDryRunResponse) {
+    rpc DryRunReplay(DryRunReplayRequest) returns (DryRunReplayResponse) {
         option (google.api.http) = {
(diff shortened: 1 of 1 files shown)


odpf/optimus/runtime_service.proto
```diff
@@ -194,6 +194,11 @@ service RuntimeService {
             get: "/v1/project/{project_name}/job/{job_name}/replay/{id}"
         };
     }
+    rpc GetReplayList(GetReplayListRequest) returns (GetReplayListResponse) {
+        option (google.api.http) = {
+            get: "/v1/project/{project_name}/replay/list"
```
kushsharma, line 199: we should make it follow REST principles, do we need `list`?

odpf/optimus/runtime_service.proto
```diff
@@ -194,6 +194,11 @@ service RuntimeService {
             get: "/v1/project/{project_name}/job/{job_name}/replay/{id}"
         };
     }
+    rpc GetReplayList(GetReplayListRequest) returns (GetReplayListResponse) {
```
kushsharma, line 197: What if we call `ListReplays`

odpf/optimus/runtime_service.proto
```diff
@@ -660,4 +665,21 @@ message RegisterJobEventRequest {
...
+}
+
+message GetReplayListResponse {
+    repeated ReplaySpec replay_list = 1;
+}
+
+message ReplaySpec {
+    string id = 1;
+    string job_name = 2;
+    string start_date = 3;
```
kushsharma, line 681: why not `google.protobuf.Timestamp`

kushsharma (changes requested):

odpf/optimus/runtime_service.proto
arinda-arif: i'm fine either way, but since start and end date is a date, i figured the response we are sending is already represented as expected, not in protobuf timestamp. any concern on this?

odpf/optimus/runtime_service.proto
arinda-arif: `ReplayList` sounds better to me, but you have concern on this? And should we change the GetReplayStatus too?

ravisuhag: are we considering replay as a resource? In that case, I guess it should be `replays` in endpoint and ListReplays in the method.

odpf/optimus/runtime_service.proto
```diff
@@ -179,24 +179,24 @@ service RuntimeService {
     }
     rpc ReplayDryRun(ReplayRequest) returns (ReplayDryRunResponse) {
         option (google.api.http) = {
-            post: "/v1/project/{project_name}/job/{job_name}/replay-dry-run"
+            post: "/v1/project/{project_name}/replay/{job_name}/dryrun"
```
ravisuhag, line 182: sinc dryrun is custom method. it can go as `:dry` 
https://cloud.google.com/apis/design/custom_methods

arinda-arif: makes sense. let me change it to `:dryrun` and test it

odpf/optimus/runtime_service.proto
arinda-arif: updated to not use list

odpf/optimus/runtime_service.proto
arinda-arif: replays seems more descriptive. but i'm seeing other endpoints are not in a plural word, for example when getting projects or resource specs. thoughts?

arinda-arif: understood why `ListReplays` is better, just went through the method naming [convention](https://cloud.google.com/apis/design/naming_convention#method_names).

arinda-arif: seems our project is using singular in all the endpoint, do you think it is acceptable to keep as it is @ravisuhag ? if not, we need to apply the changes to the others as well to keep it in uniform. cc @kushsharma

odpf/optimus/runtime_service.proto
kushsharma: @ravisuhag using colon for identifying a custom verb is one way of doing it, other way is to split these custom verbs via slash. The advantage we have here when not using `:verb` is it avoid conflicts with the resource name. For example
```
/v1/project/{project_name}/replay/{job_name}/dryrun
/v1/project/{project_name}/replay/{job_name}:dryrun
```

Also I wonder how will the gateway regex be able to differentiate that when a URL looks like
```
/v1/project/test-project/replay/test-job:dryrun
```
job_name is `test-job:dryrun` or `test-job`? So these colon based verbs are nice when we have a fixed endpoint for example like
```
/v1/project/{project_name}/replay:dryrun
```
and have everything in the POST payload but then we won't be able to do GET calls nicely.

Its very rare we will be going to allow a job name that contains a `/` whereas a job name can contain `:` now or in the future. These colon based verbs can be used if we split them after a slash though like
```
/v1/project/{project_name}/replay/{job_name}/:dryrun
```
But then this looks like more of a choice of convention. For [example](https://docs.github.com/en/rest/reference/gists#star-a-gist) github follows without colon custom verbs. What do you think?

ravisuhag: Ok cool. fine with slash as well.

odpf/optimus/runtime_service.proto
arinda-arif: Planning to resolve this and thinking to just follow the current Optimus convention to use singular, and if a change in convention is needed, will handle it in a different issue. Please let me know if you have any concern on this, @ravisuhag

odpf/optimus/runtime_service.proto
arinda-arif: discussed this and decided to use timestamp

kushsharma (approved):

PR #26, opened 2021-08-24
refactor(optimus): update misconfigured api path for runtime_service.proto

scortier:
### Description 
For Optimus : 
1. Updated the misconfigured api path 
2. Adding host entry with the default port in which server starts.

### DIFF
--- a/odpf/optimus/runtime_service.proto
+++ b/odpf/optimus/runtime_service.proto
@@ -20,7 +20,9 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
     external_docs: {
         description: "Optimus server";
     }
-    schemes: HTTP;
+    schemes: HTTP; 
+    host: "127.0.0.1:9100";
+    base_path: "/api";
 };
 
 // WARNING: This is still in active development and can have breaking changes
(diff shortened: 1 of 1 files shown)


rohilsurana:
Looks like the previous import path is available form compilation, but the one in new change isn't.

odpf/optimus/runtime_service.proto
```diff
@@ -21,6 +21,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
         description: "Optimus server";
     }
     schemes: HTTP;
+    hosts: "localhost:9100/api";
```
ravisuhag, line 24: @scortier @kushsharma What is the need of this?

odpf/optimus/runtime_service.proto
```diff
@@ -159,18 +160,21 @@ service RuntimeService {
...
-    // Datastore CRUD
+    //Database CRUD operations
+    // CreateResource registers a new resource of a namespace which belongs to a project
     rpc CreateResource(CreateResourceRequest) returns (CreateResourceResponse) {
         option (google.api.http) = {
             post: "/v1/project/{project_name}/namespace/{namespace}/datastore/{datastore_name}/resource"
             body: "*"
         };
     }
+   //ReadResource reads a provided resource spec of a namespace
```
ravisuhag, line 171: @scortier Space after commnet.

rohilsurana:
@scortier rebased it from main.

odpf/optimus/runtime_service.proto
```diff
@@ -21,6 +21,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
         description: "Optimus server";
     }
     schemes: HTTP;
+    host: "localhost:9100/api";
```
rohilsurana, line 24: Originally commented by @ravisuhag : @scortier @kushsharma What is the need of this?
(Had rebased it to fix pipeline)

odpf/optimus/runtime_service.proto
scortier: While using any api from optimus locally , 
1. The curl path is not correct ,so have to change it every time while using it.
2. The port is only to sync up with the optimus server , same as in `.optimus.yaml`.

rohilsurana: @scortier aren't both same binaries ? Can't we set the defaults in same place for both in the codebase itself?

kushsharma: There are two ways we can sort this:
1. Adding `host` entry with the default port optimus server starts.
2. Parsing json file in optimus server, find out which port optimus is running, add json `host` field with the provided host, serialize it back and host it on some optimus route.

At the moment, optimus doesn't host swagger file on server endpoint but when we do host, I think 2nd point could be used to avoid polluting the proto file but until that is done I think 1st approach should be fine so that even if they do run this swagger file locally, swagger UI works?

Also it should be `127.0.0.1:9100/api`

rohilsurana: 👍 
We should move `/api` to `base_path` field.

PR #28, opened 2021-08-30
feat(optimus): add flag to skip optimus assets compilation on CompileAssetsResponse

vianhazman:
Prerequisite for https://github.com/odpf/optimus/issues/74

### DIFF
--- a/odpf/optimus/plugins/cli.proto
+++ b/odpf/optimus/plugins/cli.proto
@@ -118,4 +118,5 @@ message CompileAssetsRequest {
 }
 message CompileAssetsResponse {
     Assets assets = 1;
+    bool skip_compile = 2;
 }


odpf/optimus/plugins/cli.proto
```diff
@@ -118,4 +118,5 @@ message CompileAssetsRequest {
 }
 message CompileAssetsResponse {
     Assets assets = 1;
+    bool skip_compile = 2;
```
kushsharma, line 121: can you add a comment here as well something like
```
// skip_compile if true, will override the default behaviour of Optimus core
// by skipping compilation of assets again if left non-templatized
```

ravisuhag:
@kushsharma @vianhazman Is this PR still needed?

kushsharma:
I don't think this is needed at the moment.

Issue #29, opened 2021-08-30
Use buf to manage external proto dependencies


rohilsurana:
**Is your feature request related to a problem? Please describe.**
As of now files from external sources are cloned and placed in `odpf/third_party` dir to be used in imports.

**Describe the solution you'd like**
Define the dependencies in buf.yaml and let buf handle it at time of generation.

PR #30, opened 2021-08-30
refactor: use buf.yaml to get third-party dependencies

rohilsurana:
Resolves https://github.com/odpf/proton/issues/29

Instead of vendoring the third-party protobuf definitions use buf to get them.


odpf/metadata/dataset.proto
```diff
@@ -1,5 +0,0 @@
-syntax = "proto3";
```
rohilsurana, line 1: @StewartJingga it seems like this was creating a problem while trying to generate go code from it. Since it is missing the `package` specifier and the `go_package` option. Is this being used somewhere as of now ?

ravisuhag: @rohilsurana This file can be completely removed.

.github/workflows/validate.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: bufbuild/buf-setup-action@v0.3.1
         with:
-          version: 0.41.0
+          version: 0.54.1
```
rohilsurana, line 14: Upraded to be able to use api version: v1 of buf.yaml

buf.yaml
```diff
@@ -1,7 +1,10 @@
...
+deps:
+  - buf.build/googleapis/googleapis
+  - buf.build/grpc-ecosystem/grpc-gateway
 breaking:
   use:
-    - WIRE
\ No newline at end of file
+    - WIRE
+  except:
+    - RESERVED_MESSAGE_NO_DELETE
```
rohilsurana, line 10: openapiv2 from grpc-gateway had some changes that are not compatible with this rule. We can remove this rule going forward.

Basically they had reserved fields so they could use them later, and now they implemented those.

ravisuhag (approved):
LGTM

kushsharma (approved):
LGTM

Issue #23, opened 2021-08-15
Proposal for proton structure 


ravisuhag:
Proton is the central repository for all protobuf definitions. Which includes APIs, entities, events used across the entire ODPF ecosystem. To make the distinctions more clear. The following structure can be used. 

```
├── odpf
│   ├── entities
│   │   ├── README.md
│   │   ├── common
│   │   │   ├── event.proto
│   │   │   ├── resource.proto
│   │   │   └── timestamp.proto
│   │   ├── facets
│   │   │   ├── columns.proto
│   │   │   ├── lineage.proto
│   │   └── resources
│   │       ├── dashboard.proto
│   │       ├── group.proto
│   ├── apis
│   │   ├── guardian
│   │   │   ├── v1
│   │   │   |    ├── guardian.proto
│   │   ├── optimus
│   │   │   ├── v1
│   │   │   |    ├── guardian.proto
  
```
This will also allow package naming to be more consistent. 
`io.odpf.apis.guardian.v1`
`io.odpf.enttiies.resources.topic`
`io.odpf.events.metadata.optimusJob`

kushsharma:
@ravisuhag can you maybe create a local directory and use `tree` to format the preview, looks a little confusing at the moment.
- How do we differentiate between `apis`, `entities` and `resources`? 
- Should we again create the same hierarchy as we have in `apis` in `resources` and `events`? Like `/odpf/events/optimus/v1`?

ravisuhag:
@kushsharma updated.

kushsharma:
Structure LGTM, although I think we can either name `entities` as just `metadata`(or something else) if that's the only thing we are using that folder for?

rohilsurana:
As per the discussion we plan to remove the `odpf` directory from the root and bring all it's subdirs to the root of the repo -
```
.
├── LICENSE
├── README.md
├── buf.lock
├── buf.yaml
├── docs
├── assets
├── guardian
|   ├── v1alpha1
│   └── v1
├── metadata
├── optimus
│   └── v1
└── raccoon
    └── v1
```
And we would also be following directory and package based api versioning.

And these would be how our different package names would look like -
**Java -**
```
io.odpf.proton.guardian.v1
io.odpf.proton.optimus.v1
```

**Proto -**
```
odpf.guardian.v1
odpf.optimus.v1
```
**go -**
```
github.com/odpf/proton/guardian/v1
github.com/odpf/proton/optimus/v1
```

kushsharma:
[...]

PR #36, opened 2021-09-02
feat(stencil): add stencil protos

harikrishnakanchi:


odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,87 @@
+syntax = "proto3";
+package odpf.stencil.v1;
+
+import "google/api/annotations.proto";
+import "google/api/field_behavior.proto";
+import "google/protobuf/descriptor.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+
+
+option go_package = "github.com/odpf/stencil/server/api/v1/pb";
```
rohilsurana, line 10: whats the reason to have `/pb` at the end ?

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,88 @@
...
+
+service StencilService {
+	rpc UploadDescriptors (UploadRequest) returns (UploadResponse) {}
+	rpc DownloadDescriptors (DownloadRequest) returns (DownloadResponse) {}
+	rpc ListSnapshots(ListSnapshotsRequest) returns (ListSnapshotsResponse) {
+		option (google.api.http) = {
+			get: "/v1/snapshots"
+		};
+	}
+	rpc UpdateLatest(UpdateLatestRequest) returns (Snapshot) {
```
rohilsurana, line 82: can we call this `UpdateSnapshot` or `UpdateLatestSnapshot` instead? To align with the googles standard method naming convention.

harikrishnakanchi: UpdateLatestSnapshot kind of misleading...It's more like MarkSnapshotAsLatest. WDYT?

rohilsurana: Lets just keep the resource we are operating on at the end. So something like `MarkLatestSnapshot`

odpf/stencil/v1/stencil.proto
rohilsurana: Lets change this to something like - `github.com/odpf/proton/stencil/v1` ?

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,86 @@
...
+
+service StencilService {
+	rpc UploadDescriptor (UploadDescriptorRequest) returns (UploadDescriptorResponse) {}
+	rpc DownloadDescriptor (DownloadDescriptorRequest) returns (DownloadDescriptorResponse) {}
+	rpc ListSnapshots(ListSnapshotsRequest) returns (ListSnapshotsResponse) {
+		option (google.api.http) = {
+			get: "/v1/snapshots"
+		};
+	}
+	rpc PromoteSnapshot(PromoteSnapshotRequest) returns (Snapshot) {
```
rohilsurana, line 81: @ravisuhag wdyt? for the GRPC endpoint name and the http api ?

odpf/stencil/v1/stencil.proto
ravisuhag: @harikrishnakanchi Yes, package name should be `github.com/odpf/proton/stencil/v1`. Will this require the server folder change is it?

harikrishnakanchi: then we have to remove odpf folder and move stencil folder to root folder.

rohilsurana: as discussed we can have `github.com/odpf/proton/stencil/v1` package but don't need to change the folder structure for now, we will do that as part of another issue already in discussion at #23

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,86 @@
...
+}
+
+message ListSnapshotsRequest {
+	string namespace = 1;
+	string name = 2;
+	string version = 3;
+	bool latest = 4;
+}
+
+message PromoteSnapshotRequest {
```
kushsharma, line 69: What does promote here mean? Promote to `release` channel? `latest` channel?

odpf/stencil/v1/stencil.proto
kushsharma: Has anyone tried using package alias using `;`? Because having version at the end of package path will generate the import paths as `v1/v2/...` which I don't think is very nice while importing generated structs. So something like
```
option go_package = "github.com/odpf/stencil/server/api/v1;servicepb";
```

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,86 @@
...
+	string version = 3 [(google.api.field_behavior) = REQUIRED];
+	bytes data = 4 [(google.api.field_behavior) = REQUIRED];
+	bool latest = 5;
+	bool dryrun = 6;
+	repeated string skiprules = 7;
+}
+
+message UploadDescriptorResponse {
+	bool success = 1;
+	bool dryrun = 2;
```
kushsharma, line 54: Can you help me understand, what are we using this field for?

odpf/stencil/v1/stencil.proto
rohilsurana: Promote to `latest`. Don't think we have anything for `release` as of now.

odpf/stencil/v1/stencil.proto
kushsharma: `returns (Snapshot)` returning a `resource` message instead of a `response` could be dangerous for backward compatibility.

harikrishnakanchi: This API updates snapshot resource. Returning updated snapshot payload makes sense. I don't see response getting changed without changing snapshot model.

odpf/stencil/v1/stencil.proto
rohilsurana: We can always import in golang like `stencil "github.com/odpf/proton/stencil/v1"` ?

odpf/stencil/v1/stencil.proto
rohilsurana: if `dryrun: true` the API just does all the validations and returns the response without actually saving the descriptor.
This will be used in CLI or client side.

odpf/stencil/v1/stencil.proto
rohilsurana: we are going with `github.com/odpf/proton/stencil/v1;stencilv1`.
Decided after looking at the buf docs for managed mode, which we might use in future - 
https://docs.buf.build/generate/managed-mode#:~:text=if%20the%20protobuf%20file's%20package%20declaration

odpf/stencil/v1/stencil.proto
rohilsurana: More of a cosmetic thing to show shell/cli users that they triggered it with the dry run flag.

odpf/stencil/v1/stencil.proto
rohilsurana: We could look into implementing channels in future. For now this just promotes to `latest`.

odpf/stencil/v1/stencil.proto
rohilsurana: Plan is to wrap this into response so it becomes easier in future to extend the API, example the channels.

rohilsurana (approved):

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,91 @@
...
+	rpc DownloadDescriptor (DownloadDescriptorRequest) returns (DownloadDescriptorResponse) {}
+	rpc ListSnapshots(ListSnapshotsRequest) returns (ListSnapshotsResponse) {
+		option (google.api.http) = {
+			get: "/v1/snapshots"
+		};
+	}
+	// PromoteSnapshot promotes particular snapshot version as latest
+	rpc PromoteSnapshot(PromoteSnapshotRequest) returns (PromoteSnapshotResponse) {
+		option (google.api.http) = {
+			patch: "/v1/snapshots/{id}:promote"
```
ravisuhag, line 88: @harikrishnakanchi @rohilsurana For now lets stick to `/promote` unless we decide to use this approach all across our services.

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,91 @@
...
+}
+
+message UploadDescriptorRequest {
+	string namespace = 1 [(google.api.field_behavior) = REQUIRED];
+	string name = 2 [(google.api.field_behavior) = REQUIRED];
+	string version = 3 [(google.api.field_behavior) = REQUIRED];
+	bytes data = 4 [(google.api.field_behavior) = REQUIRED];
+	bool latest = 5;
+	bool dryrun = 6;
+	repeated string skiprules = 7;
```
ravisuhag, line 49: @harikrishnakanchi @rohilsurana Rules can go as a message in itself. Which can have sub fields `include` and `exclude`

odpf/stencil/v1/stencil.proto
```diff
@@ -0,0 +1,91 @@
...
+
+message PromoteSnapshotRequest {
+	int64 id = 1;
+}
+
+message PromoteSnapshotResponse {
+	Snapshot snapshot = 1;
+}
+
+service StencilService {
```
ravisuhag, line 77: @harikrishnakanchi Lets move service definition to top of file.

odpf/stencil/v1/stencil.proto
harikrishnakanchi: Adding Rules as Enum.
```
enum Rules {
	FILE_NO_BREAKING_CHANGE = 0;
	MESSAGE_NO_DELETE = 1;
	FIELD_NO_BREAKING_CHANGE = 2;
	ENUM_NO_BREAKING_CHANGE = 3;
}
```

ravisuhag: @harikrishnakanchi With this as well, we would still need include and exclude as part of a wrapper messages, right.

ravisuhag:
```
...
Checks checks =7; 
}
message Checks {
  repeated string use = 1;
  repeated string exclude = 2;
}

PR #38, opened 2021-09-03
feat: add common service with version endpoint to expose version and other info of the server

rohilsurana:
The purpose of this service is to act as a common way to expose version and build info of our services.
This will also house other common endpoints such as one to get server configurations.

### DIFF
--- a/odpf/common/v1/service.proto
+++ b/odpf/common/v1/service.proto
@@ -14,7 +14,7 @@ option java_outer_classname = "CommonServiceProto";
 
 option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
     info: {
-        version: "0.1";
+        version: "0.1.0";
     };
     external_docs: {
         description: "Common endpoints for all services";
@@ -25,22 +25,25 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
 service CommonService {
     rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {
         option (google.api.http) = {
-            get: "/v1/version"
+            post: "/v1/version"
+            body: "*"
         };
     }
 }
 
-message GetVersionRequest {}
+message GetVersionRequest {
+    Version client = 1;
+}
 
-// Keeping a response envelope over version to add more info in future like latest: bool, latestVersion: Version etc
 message GetVersionResponse {
-    Version version = 1;
+    Version server = 1;
 }
 
 message Version {
     string version = 1;
-    string commit_id = 2;
+    string commit = 2;
     google.protobuf.Timestamp build_time = 3;
     string lang_version = 4;
-    string os_arch = 5;
+    string os = 5;
+    string architechture = 6;
 }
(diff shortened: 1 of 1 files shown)


odpf/common/v1/service.proto
```diff
@@ -0,0 +1,46 @@
...
+    rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {
+        option (google.api.http) = {
+            get: "/v1/version"
+        };
+    }
+}
+
+message GetVersionRequest {}
+
+// Keeping a response envelope over version to add more info in future like latest: bool, latestVersion: Version etc
```
kushsharma, line 35: no need for this comment I think, we should always be making our `rpc` by wrapping them in request/response wrappers anyway.

odpf/common/v1/service.proto
```diff
@@ -0,0 +1,46 @@
...
+
+service CommonService {
+    rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {
+        option (google.api.http) = {
+            get: "/v1/version"
+        };
+    }
+}
+
+message GetVersionRequest {}
```
kushsharma, line 33: should we also think of passing the client version here? So that the server can log what all client versions are interacting with the service? I do this in optimus.

rohilsurana: Sounds good, will make the change

rohilsurana:
@ravisuhag @AkarshSatija are we ok with service and package name `common` here ?

odpf/common/v1/service.proto
```diff
@@ -0,0 +1,49 @@
...
+    Version server = 1;
+}
+
+message Version {
+    string version = 1;
+    string commit = 2;
+    google.protobuf.Timestamp build_time = 3;
+    string lang_version = 4;
+    string os = 5;
+    string architechture = 6;
```
kushsharma, line 48: `architecture` typo

ravisuhag:
> @ravisuhag @AkarshSatija are we ok with service and package name `common` here ?

Another one would be core?

odpf/common/v1/service.proto
rohilsurana: Fixed! 😅

kushsharma:
LGTM 👍 

I'll prefer `common` over `core`. `core` keyword looks a little restrictive and internal to the system and `common` looks like a utility package that is welcoming to all.

kushsharma (approved):

odpf/common/v1/service.proto
```diff
@@ -0,0 +1,49 @@
...
+message GetVersionResponse {
+    Version server = 1;
+}
+
+message Version {
+    string version = 1;
+    string commit = 2;
+    google.protobuf.Timestamp build_time = 3;
+    string lang_version = 4;
+    string os = 5;
```
rohilsurana, line 47: what do we think about making some of these enums? Like OS and ARCH ?

ravisuhag: That will be better.

ravisuhag (approved):
LGTM

odpf/common/v1/service.proto
rohilsurana: Was checking for different languages have different ways to denote os and arch, everyone will need to maintain mappings for the same. Hence letting these be of string type.

PR #40, opened 2021-09-06
feat(optimus): add dryrun backup api path, request response model

arinda-arif:
Adding:
- BackupDryRun api path
- BackupDryRunRequest model
- BackupDryRunResponse model
to support dry run backup feature in Optimus


arinda-arif:
> Why are files like `odpf/optimus/cluster/command.proto` and `odpf/common/v1/service.proto` shown as updated? I think the PR is not properly rebased?

yes 👍  i have rebased it just now

kushsharma:
LGTM

PR #43, opened 2021-09-18
feat: add details to each resource of an appeal

ishanarya0:
Added a details field to each of the resources of an appeal.

### DIFF
--- a/odpf/guardian/guardian.proto
+++ b/odpf/guardian/guardian.proto
@@ -23,7 +23,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
 };
 
 // WARNING: This is still in active development and can have breaking changes
-service Guardian {
+service GuardianService {
     rpc ListProviders(ListProvidersRequest) returns (ListProvidersResponse) {
         option (google.api.http) = {
             get: "/providers"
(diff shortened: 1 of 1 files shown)


ishanarya0:
Changed the service name from Guardian to GuardianService, as in current use by Guardian repo.

https://github.com/odpf/guardian/blob/9a4200f173b66b18261b79be31f8b09407927fe8/api/proto/odpf/guardian/guardian.pb.gw.go#L34

odpf/guardian/guardian.proto
```diff
@@ -23,7 +23,7 @@ option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
 };
 
 // WARNING: This is still in active development and can have breaking changes
-service Guardian {
+service GuardianService {
```
rahmatrhd, line 26: let's keep it as `Guardian`

rahmatrhd (changes requested):

odpf/guardian/guardian.proto
ishanarya0: Sure, but then we need to update it in the Guardian repo as well.

rahmatrhd: yes, will update from an upcoming PR

rahmatrhd: cancelling this suggestion as per this [previous comment](https://github.com/odpf/proton/pull/17#discussion_r670766457)

rahmatrhd (approved):

PR #44, opened 2021-09-23
feat(optimus): add backup api path, request response model

arinda-arif:


odpf/optimus/runtime_service.proto
```diff
@@ -750,3 +761,34 @@ message BackupDryRunRequest {
...
+}
+
+message ListBackupsRequest {
+    string project_name = 1;
+    string datastore_name = 2;
+    string namespace = 3;
+}
+
+message ListBackupsResponse {
+    repeated BackupSpec backup_list = 1;
```
kushsharma, line 786: I think we can call this `backups` instead of backup_list.

odpf/optimus/runtime_service.proto
```diff
@@ -210,6 +210,17 @@ service RuntimeService {
             body: "*"
         };
     }
+    rpc Backup(BackupRequest) returns (BackupResponse) {
+        option (google.api.http) = {
+            post: "/v1/project/{project_name}/namespace/{namespace}/datastore/{datastore_name}/backup"
+            body: "*"
+        };
+    }
+    rpc ListBackups(ListBackupsRequest) returns (ListBackupsResponse) {
```
kushsharma, line 219: I think we should call this `ListBackup`

odpf/optimus/runtime_service.proto
arinda-arif: agree.

odpf/optimus/runtime_service.proto
```diff
@@ -750,3 +761,34 @@ message BackupDryRunRequest {
...
+    string project_name = 1;
+    string datastore_name = 2;
+    string namespace = 3;
+}
+
+message ListBackupsResponse {
+    repeated BackupSpec backup_list = 1;
+}
+
+message BackupSpec {
```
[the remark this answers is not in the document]
arinda-arif: yes, because BackupSpec is only being used on ListBackups response and currently we are not showing anything config-related.

PR #45, opened 2021-09-24
feat: add shield proto

krtkvrm:
Checklist
- [ ]  Add Validation
- [ ] Activity Endpoint
- [ ] Org Proto
- [ ] Project Proto


rohilsurana:
Hey can we follow the guidelines as mentioned here - https://cloud.google.com/apis/design/standard_methods

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,496 @@
...
+  info: {
+    title: "Shield",
+    version: "0.1.0";
+  };
+  schemes: HTTP;
+};
+
+service Shield {
+  // Users
+  rpc GetAllUsers(GetAllUsersRequest) returns (GetAllUsersResponse) {
```
kushsharma, line 26: We can get rid of `All` usage in rpc names. Just `GetUsers ` or `GetGroups`

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,496 @@
...
+message Group {
+  string id = 1;
+  string name = 2 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
+  string slug = 3;
+  repeated Admin admins = 4;
+  bool isMember = 5;
+  int32 memberCount = 6;
+  string orgId = 7;
+  google.protobuf.Struct metadata = 8;
+  google.protobuf.Timestamp createdAt = 9;
```
kushsharma, line 341: use field names as `created_at ` or `is_member `

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,496 @@
...
+
+message CreateRoleBody {
+  string name = 1 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
+  repeated string tags = 2;
+  repeated string actions = 3;
+  google.protobuf.Struct metadata = 4;
+}
+
+message CreateRoleRequest {
+  CreateRoleBody body = 1;
```
kushsharma, line 394: If this `CreateRoleBody` is tightly part of `CreateRoleRequest` why not we make it a sub-message of this request? If we don't think we want to tightly couple them then just embed these fields directly to `CreateRoleRequest`?

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,496 @@
...
+}
+
+message GetGroupRequest {
+  string id = 1;
+}
+
+message GetAllGroupsRequest {
+  string userId = 1;
+//  string orgId = 2;
+//  string org = 3;
```
kushsharma, line 369: do we need to clean this up?

odpf/shield/v1/shield.proto
rsbh: `CreateRoleBody` will also be used by `UpdateRoleRequest`.

rohilsurana: Can't we utilize the `message Role` itself instead of `CreateRoleBody` ?

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,546 @@
...
+  info: {
+    title: "Shield",
+    version: "0.1.0";
+  };
+  schemes: HTTP;
+};
+
+service ShieldService {
+  // Users
+  rpc GetUsers(GetUsersRequest) returns (GetUsersResponse) {
```
rohilsurana, line 26: Let's call this `ListUsers`

odpf/shield/v1/shield.proto
```diff
@@ -0,0 +1,546 @@
...
+      post: "/v1/users",
+      body: "body";
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      tags: "User";
+      summary: "Create User";
+    };
+  }
+
+  rpc GetUserByID(GetUserByIDRequest) returns (GetUserByIDResponse) {
```
rohilsurana, line 46: Let's call this `GetUser`

rohilsurana: The reason I am asking is that we should be able to utilize the same method in future for other GetUser requests as well on the basis of field that is set.
And if required we can add aliases for the http endpoints.

PR #46, opened 2021-09-27
feat: remove (soft) resources from db for deleted actual resources in the provider

ishanarya0:
Changes with respect to this feature: https://github.com/odpf/guardian/pull/54


odpf/guardian/guardian.proto
```diff
@@ -72,7 +72,7 @@ service GuardianService {
 
     rpc ListResources(ListResourcesRequest) returns (ListResourcesResponse) {
         option (google.api.http) = {
-            get: "/resources"
+            get: "/resources/{is_deleted}"
```
rahmatrhd, line 75:
```suggestion
            get: "/resources"
```

PR #47, opened 2021-09-30
feat: add stencil search API

harikrishnakanchi:


odpf/stencil/v1/stencil.proto
```diff
@@ -89,15 +94,38 @@ message UploadDescriptorResponse {
...
+
+message SearchResult {
+	message File {
+		string path = 1;
+		string package = 2;
+		repeated string messages = 3;
+		repeated string fields = 4;
+	}
+	Snapshot snapshot = 1;
+	repeated File files = 5;
```
kushsharma, line 124: any reason we choose index 5 straight?

harikrishnakanchi: Missed updating field number. we had couple of other fields in between before. Updated now.

PR #55, opened 2021-11-02
fix: organization body in group

krtkvrm:
- Adds org_id while creating the group
- Adds Organization body while fetching the group(s)
- Make org_id obselete

### DIFF
--- a/odpf/shield/v1/shield.proto
+++ b/odpf/shield/v1/shield.proto
@@ -341,19 +341,15 @@ message CreateGroupRequest {
 }
 
 message Group {
-  reserved 6;
-  reserved "org_id";
-
   string id = 1;
   string name = 2 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
   string slug = 3;
   repeated User admins = 4;
   int32 member_count = 5;
-  // string OBSOLETE_org_id = 6;  [OBSOLETE]
+  string org_id = 6;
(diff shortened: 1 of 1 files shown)


odpf/shield/v1/shield.proto
```diff
@@ -333,22 +333,27 @@ message GroupRequestBody {
...
+  reserved 6;
+  reserved "org_id";
+
   string id = 1;
   string name = 2 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
   string slug = 3;
   repeated User admins = 4;
   int32 member_count = 5;
-  string org_id = 6;
+  // string OBSOLETE_org_id = 6;  [OBSOLETE]
```
ravisuhag, line 352: @krtkvrm Since the shield GRPC version is not yet shipped. I think it is ok to make a breaking change and not reserve it.

krtkvrm: Had to add it as CI was showing Breaking Changes when I just replaced it with `Organization organization = 6`
Ref: https://github.com/odpf/proton/runs/4075364158?check_suite_focus=true#step:4:10

ravisuhag: We can merge it irrespective of that. Lets also see if we can enable disable breaking change ci specific to certain projects.

AkarshSatija (approved):

PR #56, opened 2021-11-07
Fix: add org_id in project Proto

krtkvrm:
- [x] Add org_id in project proto, that is to be associated with project
- [x] Remove admins list and member_count from Group proto


krtkvrm:
The validation is breaking due to re-numbering the fields in proto

AkarshSatija (approved):

PR #59, opened 2021-11-08
feat(assets): update assets metadata

StewartJingga:
- [x] move description to Resource
- [x] add lineage to Dashboard

### DIFF
--- a/odpf/assets/bucket.proto
+++ b/odpf/assets/bucket.proto
@@ -9,7 +9,6 @@ option go_package = "github.com/odpf/proton/assets";
 import "google/protobuf/timestamp.proto";
 
 import "odpf/assets/facets/ownership.proto";
-import "odpf/assets/facets/lineage.proto";
 import "odpf/assets/facets/properties.proto";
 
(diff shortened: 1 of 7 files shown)


odpf/assets/bucket.proto
```diff
@@ -66,19 +65,15 @@ message Blob {
...
   string source = 3;
 
-  // The description of the blob.
-  // Example: `This is a config file for x app`
-  string description = 4;
-
   // The length of the object content.
   // Example: `300`
-  int64 size = 5;
+  int64 size = 4;
```
mabdh, line 70: This will break if this one is currently used. Is it fine?

StewartJingga: right now nothing is using it and yes it is a breaking changes.

Currently this proto is only being used by Meteor only and still changing quite frequently.
For now we can still afford breaking changes until some other services start using it.

mabdh: noted.

mabdh (approved):

PR #60, opened 2021-11-11
fix(guardian)!: change Policy_IAM type

rahmatrhd:


odpf/guardian/v1beta1/guardian.proto
```diff
@@ -448,7 +448,12 @@ message Policy {
         repeated AdditionalAppeal appeals = 2;
     }
     repeated Requirement requirements = 8;
-    google.protobuf.Struct iam = 9;
+
+    message IAM {
+        string provider = 1;
+        google.protobuf.Value config = 2;
```
StewartJingga, line 454: if it is always going to be a map, we could use `Struct` instead of `Value`. wdyt?

PR #61, opened 2021-11-16
fix: correct schema for role according to proto

krtkvrm:
- [x] Adding types and definition_level for role as being defined in table

| Column name      | type     | Example                  |
|------------------|----------|--------------------------|
| id <PK>          | string   | org_manager              |
| name             | string   | Org Manager              |
| types            | []string | ["user" + "group"]       |
| namespace | string   | organization             |
| metadata         | json     | {"resource": "firehose"} |

- [x] Adding id while creating the role

### DIFF
--- a/odpf/shield/v1/shield.proto
+++ b/odpf/shield/v1/shield.proto
@@ -395,7 +395,7 @@ message Role {
   string id = 1;
   string name = 2 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
   repeated string types = 3;
-  string definition_level = 4;
+  string namespace = 4;
   google.protobuf.Struct metadata = 5;
   google.protobuf.Timestamp created_at = 6;
   google.protobuf.Timestamp updated_at = 7;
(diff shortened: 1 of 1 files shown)


krtkvrm:
The validation is breaking due to re-numbering the fields in proto

odpf/shield/v1/shield.proto
```diff
@@ -394,18 +394,19 @@ message ListGroupsResponse {
 message Role {
   string id = 1;
   string name = 2 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
-  repeated string tags = 3;
-  repeated string actions = 4;
+  repeated string types = 3;
+  string definition_level = 4;
```
AkarshSatija, line 398: Can you help me understand the use case of `definition_level`?

rsbh: It tells the policy namespace in which the role is applied. We can filter roles according to it.
We will also use it in dynamic schema generation.
I think "namespace" will be better key name

krtkvrm: yup

PR #63, opened 2021-11-16
feat(optimus): support multiple namespaces on replay and backup

arinda-arif:


odpf/optimus/runtime_service.proto
```diff
@@ -680,6 +680,8 @@ message ReplayRequest {
     string start_date = 4;
     string end_date = 5;
     bool force = 6;
+    bool ignore_downstream = 7;
+    string allowed_downstream = 8;
 }
 
 message ReplayResponse {
```
sravankorumilli, line 687: Let's add ignored_jobs in the replay response too.

odpf/optimus/runtime_service.proto
```diff
@@ -788,6 +796,7 @@ message BackupRequest {
     string description = 5;
     bool ignore_downstream = 6;
     map<string, string> config = 7;
+    string allowed_downstream = 8;
 }
 
 message BackupResponse {
```
sravankorumilli, line 802: Even backup response needs ignored resources

odpf/optimus/runtime_service.proto
arinda-arif: added ignored_jobs

odpf/optimus/runtime_service.proto
arinda-arif: added ignored_resources

odpf/optimus/runtime_service.proto
```diff
@@ -680,10 +680,13 @@ message ReplayRequest {
     string start_date = 4;
     string end_date = 5;
     bool force = 6;
+    bool ignore_downstream = 7;
+    string allowed_downstream = 8;
```
kushsharma, line 684: is it the namespace name? Can it only be one? Is it a delimited string?

arinda-arif: for now, the possible value is either: '*' or empty. but in the future if we decided to allow replaying downstream for some certain namespace (not all), we can put it a delimited string. wdyt?

kushsharma: if `allowed_downstream` is empty, isn't it the same as `ignore_downstream`?

arinda-arif: as discussed, updated to deprecate ignore_downstream and make allowed_downstream as repeated string.

odpf/optimus/runtime_service.proto
```diff
@@ -773,11 +778,13 @@ message BackupDryRunRequest {
     string resource_name = 3;
     string namespace = 4;
     string description = 5;
-    bool ignore_downstream = 6;
+    bool ignore_downstream = 6 [deprecated=true];
+    repeated string allowed_downstream = 7;
```
kushsharma, line 782: what if we rename it to `downstream_namespace_filter`? Also add some comments about how this work?

arinda-arif: sure i can add the comment. for the name, does the concern of naming it `allowed_downstream` is because it is not clear enough that it is representing namespaces? if yes, does `allowed_downstream_namespace` sounds better than `downstream_namespace_filter`? to avoid questions on what is this filter, why it is needed.

arinda-arif: as discussed, will go with `allowed_downstream_namespaces`

arinda-arif: hi @kushsharma , have changed to `allowed_downstream_namespaces` and added the comments. please take a look.

kushsharma:
LGTM

PR #64, opened 2021-11-17
feat(entropy): add entropy grpc service definition

rohilsurana:


odpf/entropy/v1beta1/service.proto
```diff
@@ -0,0 +1,98 @@
...
+  string parent = 4;
+  google.protobuf.Value configs = 5;
+}
+
+message Module {
+  string id = 1;
+}
+
+message ListResourcesRequest {
+  string parent = 1;
```
gauravsinghania, line 62: doesn't resource here makes sense? Will allow us to search based on any key?

odpf/entropy/v1beta1/service.proto
```diff
@@ -0,0 +1,98 @@
...
+
+message UpdateResourceResponse {
+  Resource resource = 1;
+}
+
+message DeleteResourceRequest {
+  string id = 1;
+}
+
+message DeleteResourceResponse {}
```
gauravsinghania, line 98: For delete, does it makes sense to send the resource that is deleted? Or just the status should suffice?

odpf/entropy/v1beta1/service.proto
rohilsurana: As of now GRPC gateway doesn't support using a FieldMask and automatically updating it based on the query params for a get request.
There is an open issue for the same, once it is available we can move to that.

odpf/entropy/v1beta1/service.proto
rohilsurana: This is standard for delete requests, but if needed we can add those in future. 
As for status, I think grpc status would suffice for now.

kushsharma: `send the resource that is deleted` how are you thinking we will leverage it? @gauravsinghania

odpf/entropy/v1beta1/service.proto
```diff
@@ -0,0 +1,100 @@
...
+  string kind = 2;
+}
+
+message ListResourcesResponse {
+  repeated Resource resources = 1;
+}
+
+message GetResourceRequest {
+  string urn = 1;
+  string kind = 2;
```
kushsharma, line 71: Do we see urn as unique? How do we use kind here? And what about id? How is that being utilized?
Same in `Delete` call as well?

rohilsurana: We plan to get rid of `id` and make `urn` as our unique identifier.

kushsharma: Okay, and what about `kind`? Are `urn` not unique enough to find the resource?

rohilsurana: Yes we will get rid of `kind` as well. At first the responsibility of generating and maintaining `urn` was with each modules package, but we have moved that into a central package allowing us to have proper constraints on it.

rohilsurana: Fixed this.

pyadav (commented):
LGTM

PR #65, opened 2021-11-22
feat(assets): add usage metrics to table profile

mabdh:
- Add several fields in `assets/table/TableProfile` to store table usage stats
  - Add `usage_count` as a metric for table usage
  - Add `common_join` as a metric for the list of joined tables
- Related [PR](https://github.com/odpf/meteor/pull/271) in meteor

### DIFF
--- a/odpf/assets/table.proto
+++ b/odpf/assets/table.proto
@@ -69,13 +69,13 @@ message TableProfile {
 
   int64 usage_count = 4;
   
-  repeated TableCommonJoin common_join = 5;
+  repeated Join joins = 5;
 
-  repeated string filter_conditions = 6;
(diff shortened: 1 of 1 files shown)


odpf/assets/table.proto
```diff
@@ -66,4 +66,19 @@ message TableProfile {
...
+
+  int64 usage_count = 4;
+  
+  repeated Join joins = 5;
+
+  repeated string filters = 6;
 }
+
+// TableCommonJoin is the metric of which are other tables that are joined with this table
+message TableCommonJoin {
```
ravisuhag, line 78: @mabdh This will also be renamed to joins right?

PR #70, opened 2021-12-03
feat: add resource configuration for job

irainia:
### Context

There's a requirement to allow resource configuration in job for Optimus. It is mentioned in [Optimus issue](https://github.com/odpf/optimus/issues/114).

### Changes

The change is only to add additional message, using it inside the job spec.


odpf/optimus/runtime_service.proto
```diff
@@ -255,6 +255,20 @@ message JobSpecHook {
...
 
+message JobSpecResourceConfig {
+    string cpu = 1;
+    string memory = 2;
+}
+
+message JobSpecResource {
+    JobSpecResourceConfig request = 1;
+    JobSpecResourceConfig limit = 2;
+}
```
kushsharma, line 266: Should we prefix them with `Metadata`? `Resource` is a term already used for something else in Optimus. It doesn't directly convey the message these are execution time resource limits/requests

irainia: Oh right. Then, are `JobSpecMetadataResource` and `JobSpecMetadataResourceConfig` good enough?

kushsharma: yeah sounds good.

irainia: it's updated as suggested

PR #83, opened 2021-12-17
feat(optimus): add namespace name to RegisterSecretRequest

arinda-arif:


odpf/optimus/runtime_service.proto
```diff
@@ -554,6 +554,8 @@ message RegisterSecretRequest {
     string project_name = 1;
     string secret_name = 2;
     string value = 3; // base64 encoded secret value
+    string namespace_name = 4;
+    bool updateOnly = 5;
```
kushsharma, line 558: this should not be camel case

kushsharma (changes requested):

odpf/optimus/runtime_service.proto
sbchaos: Thank you for pointing out. I will make this change.

odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -559,13 +568,26 @@ message RegisterSecretRequest {
...
+message UpdateSecretRequest {
+    string project_name = 1;
+    string secret_name = 2;
+    string value = 3; // base64 encoded secret value
+    string namespace_name = 4;
+}
+
+message UpdateSecretResponse {
+    bool success = 1;
+    string message = 2;
```
kushsharma, line 588: How are we using these? I don't think we need a success flag here?

odpf/optimus/runtime_service.proto
sbchaos: Marking as resolved, changes already done.

odpf/optimus/core/v1beta1/runtime.proto
sbchaos: Sure, will remove `success` flag from the proto in next PR.

PR #86, opened 2022-01-11
feat(shield): add admins api

rsbh:


odpf/shield/v1beta1/shield.proto
```diff
@@ -605,6 +712,66 @@ message ListGroupsResponse {
...
+  repeated User users = 1;
+}
+
+message RemoveGroupAdminRequest {
+  string id = 1;
+  string user_id = 2;
+}
+
+message RemoveGroupAdminResponse {
+  string message = 1;
```
krtkvrm, line 772: will this return something like "{User} removed from {Group}"

PR #88, opened 2022-01-11
feat(optimus): add api for listing the secrets

sbchaos:


odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -559,13 +575,40 @@ message RegisterSecretRequest {
...
+    bool success = 1;
+    string message = 2;
+}
+
+message ListSecretsRequest {
+    string project_name = 1;
+}
+
+message ListSecretsResponse {
+    message Secret {
```
kushsharma, line 603: I don't see we have a `Get` call for secrets yet. I guess we will have it soon? And when we do are we going to create a separate message for it?

sbchaos: The secret message will be different in that case, that will require name and value of secret.

sbchaos:
Conflicts from previous branch, closing PR. Will reopen after resolving conflicts.

PR #92, opened 2022-01-21
feat(stencil): add support for search API

ramey:
- add RPC call `Search`
- define messages `SearchRequest`, `SearchResponse` and `SearchMeta`

### DIFF
--- a/odpf/stencil/v1beta1/stencil.proto
+++ b/odpf/stencil/v1beta1/stencil.proto
@@ -309,9 +309,12 @@ message DeleteVersionResponse {
 message SearchRequest {
   string namespace_id = 1;
   string schema_id = 2;
-  int32 version_id = 3;
-  string query = 4;
-  bool all = 5;
+  string query = 3;
+  
+  oneof version {
+    bool history = 4;
+    int32 version_id = 5;
+  }
 }
 
 message SearchResponse {
--- a/odpf/stencil/v1beta1/stencil.proto
+++ b/odpf/stencil/v1beta1/stencil.proto
@@ -127,6 +127,16 @@ service StencilService {
       summary: "Delete specified version of the schema";
     };
   }
+
+  rpc SearchSchemas(SearchSchemasRequest) returns (SearchSchemasResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/namespaces/{namespace_id}/schemas/_search"
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      tags: "schema";
+      summary: "Search all the schemas in a given namespace";
+    };
+  }
 }
 
 message Namespace {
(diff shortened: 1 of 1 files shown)


odpf/stencil/v1beta1/stencil.proto
```diff
@@ -295,3 +305,21 @@ message DeleteVersionRequest {
 message DeleteVersionResponse {
   string message = 1;
 }
+
+message SearchSchemasRequest {
+  string namespace_id = 1;
+  int32 version_id = 2;
```
ravisuhag, line 311: @ramey Version alone will not make sense here, right? The stencil will have many schemes and each schema will have different versions. So we if want to search specifically then we will need schema id as well. 
IMO, Specificity of search will work in following manner:   namespace > schema > version

ramey: Makes sense @ravisuhag, will remove version.
In that case shall we also support  `/namespace/{namespace_id}/schemas/{schema_id}/_search` and `/namespace/{namespace_id}/schemas/{schema_id}/versions/{version_id}/_search` APIs?

ravisuhag: @ramey I think we should support a `/search` API which by default will search across all namespaces, schemas, and versions(latest by default). And it will also take query params for namespace, schema, and version to filter the result down. WDYT?

ramey: Done changes as discussed.

odpf/stencil/v1beta1/stencil.proto
```diff
@@ -306,20 +306,27 @@ message DeleteVersionResponse {
...
 
-message SearchSchemasRequest {
+message SearchRequest {
   string namespace_id = 1;
-  int32 version_id = 2;
-  string query = 3;
+  string schema_id = 2;
+  int32 version_id = 3;
+  string query = 4;
+  bool all = 5;
```
ravisuhag, line 314: @ramey What is the role of all here?

ramey: `all` returns matches from all the versions, by default only the matches from the latest versions are returned

ravisuhag: Make sense. Is there some better name we can think of for this? At first look I got confused that it is related to pagination,

ramey: how does `scan_all_versions` sound?

harikrishnakanchi: what we need is oneof between `version_id` and `all` fields. User can't set both fields together right? If they do, which field should take precendence?

ramey: If both are specified then `version_id` will get the precedence.

ravisuhag: Hmm, we can think of using oneof to make it more explicit. 
Also what if we use latest as param instead of all. which by default is true? latest makes it more explicit? WDYT?

ramey: The problem is while processing the value in the application. In case no value has been specified as it is a `bool` it defaults to `false` and there is no way to find out whether it is explicitly given by the user or default `bool` value. 
Agreed with `oneof`, will change that.

ravisuhag: make sense. In that case, another name of all could be history as well.

ravisuhag: @ramey @harikrishnakanchi lets go with history then?

ramey: @ravisuhag I am planning to go with `history`

ravisuhag: @ramey cool.

ravisuhag (approved):

PR #93, opened 2022-01-25
feat(optimus): add listSecrets api, remove unused fields

sbchaos:


odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -583,9 +587,20 @@ message UpdateSecretRequest {
...
-    bool success = 1;
-    string message = 2;
+message UpdateSecretResponse {}
+
+message ListSecretsRequest {
+    string project_name = 1;
+}
+
+message ListSecretsResponse {
+    message Secret {
```
kushsharma, line 597: Shouldn't we extract this message at root level so we don't need to create again for `Get` call? We can add a value field as well but return it empty for `List` call? What do you think?

sbchaos: As of now we don't have a GET call, we can extract this message when we have the GET and add required fields.

kushsharma: That will require us to refractor the code again to use a different proto message right? Why not have it designed like that from the start?

sbchaos: We do not have any GET call planned to expose cleartext secrets over api.

sravankorumilli: Yes, get on a secret isn't planned, and  the expectation from read  is to return the plaintext secret, so the message will change again. So, we can keep it under SecretList

sravankorumilli (approved):

sravankorumilli (approved):

PR #97, opened 2022-02-02
fix(entropy)!: use enum for resource status and change UpdateResourceRequest to include only urn and configs.

rohilsurana:
BREAKING CHANGE: 
 - make Resource.status an enum type.
 - Remove resource from UpdateResourceRequest and add urn and configs only.


rohilsurana:
Breaking CI workflows can be ignored as the commit includes breaking changes on recently merged protobuf definition.

PR #98, opened 2022-02-10
Proposal - 1 : add external http sensor at optimus job dependency

siddhanta-rath:
- https://github.com/odpf/optimus/issues/138

### DIFF
--- a/odpf/optimus/core/v1beta1/runtime.proto
+++ b/odpf/optimus/core/v1beta1/runtime.proto
@@ -873,7 +873,10 @@ message GetBackupResponse {
 }
 
 message HttpSensor {
-    string url = 1;
-    map<string, string> headers = 2;
-    map<string, string> request_params = 3;
+    string name = 1;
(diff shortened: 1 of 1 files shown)


odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -870,3 +871,10 @@ message GetBackupResponse {
     BackupSpec spec = 1;
     repeated string urn = 2;
 }
+
+message HttpSensor {
+    string name = 1;
+    string url = 2;
+    map<string, string> headers = 3;
+    map<string, string> request_params = 4;
```
sravankorumilli, line 879: can be just params

sravankorumilli (approved):

odpf/optimus/core/v1beta1/runtime.proto
siddhanta-rath: addressed with Latest Commit

PR #104, opened 2022-02-21
feat(optimus): add ability to delete secret for a project

sbchaos:


odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -603,6 +610,13 @@ message ListSecretsResponse {
     repeated Secret secrets = 1;
 }
 
+message DeleteSecretRequest {
+    string project_name = 1;
+    string secret_name = 2;
```
sravankorumilli, line 615: Should take namespace details as well.
Helps in doing the authorization at proxy level.

sbchaos: Done

PR #105, opened 2022-03-03
feat: add is_deleted as param & in schema

ishanarya0:
- Add is_deleted to resource schema, in order to distinguish deleted(soft) resources.
- Add is_deleted param in list-resource request, to view deleted(soft) resources as well. By default, it shall not return deleted resources.


ishanarya0:
As discussed, we don't need to return deleted resources, so these changes won't be required as of now.

PR #107, opened 2022-03-10
feat(optimus): fetch JobRuns for given date range

siddhanta-rath:
Implementation: 
https://github.com/odpf/optimus/issues/198

### DIFF
--- a/odpf/optimus/core/v1beta1/runtime.proto
+++ b/odpf/optimus/core/v1beta1/runtime.proto
@@ -159,6 +159,14 @@ service RuntimeService {
             get: "/v1beta1/project/{project_name}/job/{job_name}/status"
         };
     }
+
+    // JobRun returns the current and past run status of jobs on a given range
+    rpc JobRun(JobRunRequest) returns (JobRunResponse) {
+        option (google.api.http) = {
+            get: "/v1beta1/project/{project_name}/job/{job_name}/run"
+        };
+    }
+
     // RegisterJobEvent notifies optimus service about an event related to job
     rpc RegisterJobEvent(RegisterJobEventRequest) returns (RegisterJobEventResponse) {
         option (google.api.http) = {
@@ -668,6 +676,23 @@ message JobStatusRequest {
 
 message JobStatusResponse {
     repeated JobStatus statuses = 1;
+}
+
+message JobRunRequest {
+    string project_name = 1;
+    string job_name = 2;
+    string start_date = 3;
+    string end_date = 4;
+    repeated string filter = 5;
+}
+
+message JobRunResponse {
+    repeated RunStatus run_status = 1;
+}
+
+message RunStatus {
+    string state = 1;
+    google.protobuf.Timestamp scheduled_at = 2;
 }
 
 message GetWindowRequest {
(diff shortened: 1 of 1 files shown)


odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -670,6 +678,23 @@ message JobStatusResponse {
...
+    google.protobuf.Timestamp start_date = 3;
+    google.protobuf.Timestamp end_date = 4;
+    repeated string filter = 5;
+}
+
+message JobRunResponse {
+    repeated RunStatus run_status = 1;
+}
+
+message RunStatus {
```
sravankorumilli, line 693: Can we use JobRun instead

sravankorumilli (changes requested):

odpf/optimus/core/v1beta1/runtime.proto
siddhanta-rath: new rpc is named as `JobRun`, is it okay to have same name for rpc and message?

sravankorumilli: I believe that should be fine, we don't have a technical limitation right

siddhanta-rath: No, I think it will make more sense to end user. as API response will look like below :
```
{
    "jobRun": [
        {
            "state": "success",
            "scheduledAt": "2022-03-01T02:00:00Z"
        },
        {
            "state": "success",
            "scheduledAt": "2022-03-02T02:00:00Z"
        }
    ]
}
```

sravankorumilli: RunStatus is more generic, we have replay runs, backup runs, job runs. Lets keep it specific.
Curious to understand how the jobRun name is coming if we are using RunStatus as the message.

sravankorumilli (changes requested):

odpf/optimus/core/v1beta1/runtime.proto
siddhanta-rath: Agree, that's why I have changed the message to` Jobrun` at the latest commit. Just gave an example earlier regarding the response structure with `jobRun`

odpf/optimus/core/v1beta1/runtime.proto
```diff
@@ -670,6 +678,23 @@ message JobStatusResponse {
...
+message JobRunRequest {
+    string project_name = 1;
+    string job_name = 2;
+    google.protobuf.Timestamp start_date = 3;
+    google.protobuf.Timestamp end_date = 4;
+    repeated string filter = 5;
+}
+
+message JobRunResponse {
+    repeated JobRun job_run = 1;
```
sravankorumilli, line 690: update to job_runs

siddhanta-rath: done

siddhanta-rath: it's proper now
```
{
    "jobRuns": [
        {
            "state": "success",
            "scheduledAt": "2022-03-01T02:00:00Z"
        },
        {
            "state": "success",
            "scheduledAt": "2022-03-02T02:00:00Z"
        }
    ]
}
```

sravankorumilli (approved):

PR #111, opened 2022-03-17
feat(optimus): change how secrets are used by plugins

sbchaos:


odpf/optimus/plugins/v1beta1/dependency_resolver.proto
```diff
@@ -21,10 +21,17 @@ service DependencyResolverModService {
     rpc GenerateDependencies(GenerateDependenciesRequest) returns (GenerateDependenciesResponse);
 }
 
+message PluginJobConfigs {
```
sravankorumilli, line 24: why do we need a segregation between secrets and regular configs, plugin implementation knows the difference

odpf/optimus/plugins/v1beta1/dependency_resolver.proto
```diff
@@ -21,10 +21,17 @@ service DependencyResolverModService {
...
+
 message GenerateDestinationRequest {
-    Configs config = 1;
-    Assets assets = 2;
-    odpf.optimus.core.v1beta1.ProjectSpecification project = 3;
+    Configs config = 1  [deprecated=true];
+    Assets assets = 2  [deprecated=true];
+    odpf.optimus.core.v1beta1.ProjectSpecification project = 3  [deprecated=true];
+
+    PluginJobConfigs job_configs = 4;
```
sravankorumilli, line 34: JobConfig & Assets we need to pass and we are passing those details, we don't need to deprecate assets.
PluginJobConfigs is not necessary imo we can leverage configs itself

odpf/optimus/plugins/v1beta1/dependency_resolver.proto
```diff
@@ -34,9 +41,11 @@ message GenerateDestinationResponse {
 }
 
 message GenerateDependenciesRequest {
-    Configs config = 1;
+    Configs config = 1  [deprecated=true];
     Assets assets = 2;
-    odpf.optimus.core.v1beta1.ProjectSpecification project = 3;
+    odpf.optimus.core.v1beta1.ProjectSpecification project = 3  [deprecated=true];
+
+    PluginJobConfigs job_configs = 4;
```
sravankorumilli, line 48: similar to the above comment

odpf/optimus/plugins/v1beta1/dependency_resolver.proto
sbchaos: Sure, will remove this

sravankorumilli (approved):

PR #113, opened 2022-03-22
Deploy multi namespace

irainia:
This change will allow for multi-namespaces deployment.

Changes:
* updating deployment of job to use streaming input
* updating deployment of resource to use streaming input


irainia:
it's confirmed by @sravankorumilli to be merged

PR #115, opened 2022-03-24
feat(compass): add asset, search, lineage, users, star, tags handlers

mabdh:


ravisuhag:
@mabdh can you please check the lint issues reported by buf on compass proto.

mabdh:
@ravisuhag yes there are some breaking changes with the main, how could we force the changes for compass proto?

ravisuhag:
> @ravisuhag yes there are some breaking changes with the main, how could we force the changes for compass proto? is it fine to merge it later with these breaking changes?

It is fine to merge with breaking changes since we are yet to integrate and release the apis. But there are som more lint issues, which also we can fix.

ravisuhag:
> @ravisuhag yes there are some breaking changes with the main, how could we force the changes for compass proto? is it fine to merge it later with these breaking changes?

It is fine to merge with these changes since we are yet to integrate and release the apis. But there are some more lint issues, which also we can fix.

mabdh:
@ravisuhag right, there is some lint errors because we re-use a message in the different rpc services, e.g. `google.protobuf.Empty`.
Sure I will fix it.

PR #122, opened 2022-04-04
feat: add provider service

ishanarya0:
- Add `Create` and `List` providers
- Add `ProviderFilters` to `Resource`

### DIFF
--- a/odpf/entropy/v1beta1/service.proto
+++ b/odpf/entropy/v1beta1/service.proto
@@ -59,6 +59,21 @@ service ResourceService {
   }
 }
 
+service ProviderService {
+  rpc ListProviders(ListProvidersRequest) returns (ListProvidersResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/providers"
+    };
+  }
+
+  rpc CreateProvider(CreateProviderRequest) returns (CreateProviderResponse) {
+    option (google.api.http) = {
+      post: "/v1beta1/providers"
+      body: "provider"
+    };
+  }
+}
+
 message Resource {
   string urn = 1;
   string name = 2;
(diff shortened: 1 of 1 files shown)


odpf/entropy/v1beta1/service.proto
```diff
@@ -76,8 +91,25 @@ message Resource {
     STATUS_COMPLETED = 5;
   }
   Status status = 7;
-  google.protobuf.Timestamp created_at = 8;
-  google.protobuf.Timestamp updated_at = 9;
+  repeated ProviderFilter provider = 8;
```
rohilsurana, line 94: Lets make this something like -
```suggestion
  repeated ProviderSelector providers = 8;
```

odpf/entropy/v1beta1/service.proto
```diff
@@ -76,8 +91,25 @@ message Resource {
...
+  google.protobuf.Timestamp created_at = 9;
+  google.protobuf.Timestamp updated_at = 10;
+}
+
+message ProviderFilter {
```
rohilsurana, line 99:
```suggestion
message ProviderSelector {
```

odpf/entropy/v1beta1/service.proto
ishanarya0: Sure!

odpf/entropy/v1beta1/service.proto
ishanarya0: Sure!

rohilsurana:
I see the CI is red as this is a breaking change. I think it should be fine since we are not using this proto in production anywhere.

rohilsurana (approved):

PR #126, opened 2022-04-06
feat(raccoon): fix buf lint issues

ramey:
Following changes are done - 

- Standardised raccoon files with respect to other odpf projects.
- Fixed lint issues

Signed-off-by: ramey <rameygirdhar@gmail.com>


ravisuhag:
@ramey Let's stick with v1beta1 for API version.

ramey:
> @ramey Let's stick with v1beta1 for API version.

@ravisuhag we already use `v1` in [raccoon](https://github.com/odpf/raccoon/blob/main/services/rest/service.go#L47). That's why made it `v1`. I think we should be consistent with both and if we make it `v1beta1` then It will be a breaking change in Raccoon.

ravisuhag:
@ramey can you illustrate where we are using v1 right now?

odpf/raccoon/v1/raccoon.proto
```diff
@@ -0,0 +1,163 @@
...
+option java_multiple_files = true;
+option java_package = "io.odpf.proton.raccoon";
+option java_outer_classname = "EventProto";
+
+option go_package = "github.com/odpf/proton/raccoon/v1;raccoonv1";
+
+
+
+service EventService {
+  rpc SendEvent (EventRequest) returns (EventResponse);
```
ramey, line 16: @ravisuhag @chakravarthyvp need your inputs here, buf lint is recommending the change of name for EventRequest and EventResponse. This will be a breaking change, so shall we ignore this or make the change? Another thing we can do is probably have both `SendEventRequest/Response` and `EventRequest/Response` with same fields for now and then slowly phase`EventRequest/Response`out. What do you folks think?

ramey:
@ravisuhag `v1` is in API path that clients use to connect to Raccoon `raccoon.gojekapi.com/v1/events`. Agree, we are introducing versioning for the first time but we already have v1 APIs. Won't that be a little confusing or it doesn't matter?

odpf/raccoon/v1/raccoon.proto
chakravarthyvp: It should not break since this changing just the names and not the tag numbers. I suggest we verify this commit in one of our integration and wire an old version proto against it.

ramey:
> I think until our APIs are not stable we should not release v1. One way to deal with raccoon hard coded path is to add alias for now. And give a timeline to EOL.

Currently we have two sets of APIs, the old `EventRequest` and `EventResponse` data models which are stable and that's why we were using `v1` and the new gRPC APIs which are not, so it makes sense for them to be under `v1beta1`.  The new gRPC API's are dependent on the old data models. So we can do two things here, either create two directories under raccon for data models and service and both have their own semantic versioning or we leave it the way it was previously and just do other field level changes? WDYT @ravisuhag

odpf/raccoon/v1/raccoon.proto
NNcrawler: The `EventRequest` is independent of GRPC and IMO we should keep it that way. However, `SendEventRequest` sounds like a good name even from a WebSocket or HTTP perspective. I don't mind renaming.

chakravarthyvp: +1 @NNcrawler 

@ramey - Are the EventRequest/Reponse/Event moved out of the service proto? I see you upvoted @NNcrawler 's comment.

We can keep the gRPC service proto as `v1beta`

ramey: No @chakravarthyvp I have kept everything in the same proto file from readability point of view. If renaming doesn't break things then let's rename them.

chakravarthyvp:
@ramey - Can we keep the `SendEventRequest/SendEvent/SendEventResponse` data models outside the raccoon.proto? These models are not coupled with gRPC and they are used by the http APIs as well. We are not generating the http paths from the service protos either.
We could refactor and merge the service/model protos when we release v1 version of the rpc protos.
Meaning, today:
odpf/raccon/ ==> All models go here
odpf/v1beta1/ ==> gRPC service proto goes here

when gRPC moves to a stable version, we would have:
/odpf/raccoon/v1/raccoon.proto ==> at this point: we will have http path generated as well here!

ravisuhag:
@chakravarthyvp Any reason to put models in v1beta1?

chakravarthyvp:
> @chakravarthyvp Any reason to put models in v1beta1?
We shouldn't put the models in v1beta1. 
I think @ramey has included this in v1beta1 as part of raccoon.proto (service proto). I'm proposing that this should be moved out of this version and be at higher level as the models are reused by http APIs & gRPC.

https://github.com/odpf/proton/pull/126#issuecomment-1110540829

Do you agree with this approach @ravisuhag ?

ravisuhag:
> > @chakravarthyvp Any reason to put models in v1beta1?
> 
> We shouldn't put the models in v1beta1. I think @ramey has included this in v1beta1 as part of raccoon.proto (service proto). I'm proposing that this should be moved out of this version and be at higher level as the models are reused by http APIs & gRPC.
> 
> [#126 (comment)](https://github.com/odpf/proton/pull/126#issuecomment-1110540829)
> 
> Do you agree with this approach @ravisuhag ?

I meant not to put in V1beta1.

chakravarthyvp:
@ravisuhag - The models (which are ideally payloads) are used by the http APIs and re-used by RPCs. This has to be agnostic of which API/RPC uses it. Clients should be able to use these models along with the API they choose and needn't tie with the RPC versions.

ravisuhag:
@chakravarthyvp v1beta1 version we want to add in the proton is not specific to just the RPC but for all proto messages actually. So, I would recommend attaching a version to models as well. We can for sure think if the versioning of models and GRPC service should be maintained separately or not.

PR #139, opened 2022-05-17
feat(optimus): add type field and deprecate ack field from DeployJobSpecificationResponse

arinda-arif:


odpf/optimus/core/v1beta1/job_spec.proto
```diff
@@ -90,10 +90,18 @@ message DeployJobSpecificationResponse {
 
   // non ack responses are more of a progress/info response
   // and not really success or failure statuses
-  bool ack = 2;
+  // this is deprecated and being replaced with type field
+  bool ack = 2 [deprecated=true];
```
sravankorumilli, line 94: want to discuss on these contracts `type` and `ack`

sravankorumilli (approved):

odpf/optimus/core/v1beta1/job_spec.proto
arinda-arif: @sravankorumilli @deryrahman  just want to confirm, is this resolved? As the PR in Optimus has merged.

PR #149, opened 2022-06-21
refactor(siren): update api request response to avoid lint error

mabdh:


odpf/siren/v1beta1/siren.proto
```diff
@@ -408,7 +407,7 @@ message ListProvidersRequest {
 }
 
 message ListProvidersResponse {
-  repeated Provider providers = 1;
```
ravisuhag, line 411: @mabdh Any reason to change this? I think better practice is to use `providers` only instead of data.

ravisuhag: Reference from googleapis (https://github.com/googleapis/googleapis/blob/c65a86879995025a01f2f3946818aa5142596a8d/google/cloud/asset/v1/asset_service.proto#L425)

ravisuhag (approved):

PR #153, opened 2022-06-30
feat(entropy): add resource history endpoint

rohilsurana:
For https://github.com/odpf/entropy/issues/70

### DIFF
--- a/odpf/entropy/v1beta1/service.proto
+++ b/odpf/entropy/v1beta1/service.proto
@@ -8,9 +8,8 @@ import "google/protobuf/timestamp.proto";
 
 option go_package = "github.com/odpf/proton/entropy/v1beta1;entropyv1beta1";
 option java_multiple_files = true;
-option java_package = "io.odpf.proton.entropy";
 option java_outer_classname = "EntropyServiceProto";
-
+option java_package = "io.odpf.proton.entropy";
 
 service ResourceService {
   rpc ListResources(ListResourcesRequest) returns (ListResourcesResponse) {
@@ -66,12 +65,12 @@ service ResourceService {
 }
 
 message ResourceDependency {
-    // Key should be as defined by the module being used for
-    // the resource.
-    string key = 1;
+  // Key should be as defined by the module being used for
+  // the resource.
+  string key = 1;
 
-    // Value should refer to an existing resource via URN.
-    string value = 2; 
+  // Value should refer to an existing resource via URN.
+  string value = 2;
 }
 
 message ResourceSpec {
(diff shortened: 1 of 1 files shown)


odpf/entropy/v1beta1/service.proto
```diff
@@ -166,3 +172,19 @@ message GetLogRequest {
...
+message ResourceHistory {
+  string urn = 1;
+  map<string, string> labels = 2;
+  google.protobuf.Timestamp created_at = 3;
+
+  google.protobuf.Value difference = 4;
+}
+
+message GetResourceHistoryRequest {
+  string urn = 1;
```
spy16, line 185: should we paginate?

odpf/entropy/v1beta1/service.proto
```diff
@@ -166,3 +172,19 @@ message GetLogRequest {
 message GetLogResponse {
   LogChunk chunk = 1;
 }
+
+message ResourceHistory {
+  string urn = 1;
+  map<string, string> labels = 2;
+  google.protobuf.Timestamp created_at = 3;
```
mabdh, line 179: Why no `updated_at`?

Is this for versioning history?

odpf/entropy/v1beta1/service.proto
rohilsurana: Left out pagination for now, we can do that on all APIs at once.

odpf/entropy/v1beta1/service.proto
rohilsurana: Yes this is for resource versioning. The thought is that the versioning should be immutable and append only.

odpf/entropy/v1beta1/service.proto
rohilsurana: Planned to add it right away, will do it for other APIs as part of this itself.

rohilsurana: Dropped pagination for now, need more details on the pagination UX.

odpf/entropy/v1beta1/service.proto
```diff
@@ -166,3 +171,20 @@ message GetLogRequest {
 message GetLogResponse {
   LogChunk chunk = 1;
 }
+
+message ResourceHistory {
```
spy16, line 175: this naming still feels wrong to me since this message itself is not the history but multiple entries of this message together form the history. 

could we name it as `ResourceSpecRevision` or something like that?

rohilsurana: We can call it `revision`, that works too.

rohilsurana: Don't want to add `Spec` to the name so it still remains a little generic in case we want to include other fields in future.

spy16: I added `Spec` because we are only maintaining history of changes to the spec. but I am okay either way.

PR #156, opened 2022-07-06
feat(shield): add limit, page, keyword for pagination and filter

niharbansal02:
Update `shield.proto`
Add `limit`, `page`, `keyword` for pagination and filter.


ravisuhag:
@niharbansal02 Let's follow the guidelines suggested in this doc for pagination.

PR #161, opened 2022-07-20
feat: simplify meteor metadata model

GrayFlash:


odpf/assets/common/v1beta2/resource.proto
```diff
@@ -0,0 +1,36 @@
...
+package odpf.assets.common.v1beta2;
+
+option java_package = "io.odpf.assets.common";
+option java_outer_classname = "ResourceProto";
+option go_package = "github.com/odpf/proton/assets/common/v1beta2;commonv1beta2";
+
+
+// Resource is a generic resource that represents a file or other resource.
+// It can be a table, job, user or group.
+message Resource {
```
GrayFlash, line 12: have flattened `Asset.Resource.*` to `Asset.*`, but could not delete the Resource.proto because it is being used in lineage.proto

odpf/assets/v1beta2/assets.proto
```diff
@@ -0,0 +1,70 @@
...
+import "odpf/assets/common/v1beta2/ownership.proto";
+import "odpf/assets/facets/v1beta2/lineage.proto";
+
+import "odpf/assets/common/v1beta2/event.proto";
+import "google/protobuf/any.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+// Job is a resource that represents a job.
+message Assets {
```
ravisuhag, line 18: @GrayFlash Lets make this singular,

odpf/assets/facets/v1beta2/lineage.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
```
ravisuhag, line 1: @GrayFlash i am thinking may be we can move all facets and common protos to a single file. thoughts? cc @StewartJingga

odpf/assets/common/v1beta2/resource.proto
StewartJingga: We can move resource to `lineage.proto` since it is the only place it is used.

odpf/assets/common/v1beta2/event.proto
```diff
@@ -0,0 +1,26 @@
...
+option java_package = "io.odpf.assets.common";
+option java_outer_classname = "EventProto";
+option go_package = "github.com/odpf/proton/assets/common/v1beta2;commonv1beta2";
+
+import "google/protobuf/timestamp.proto";
+
+// Event represents an event in the system.
+// Event is majorly used to represent the state of the system in the form of events.
+// It can be used in any schema which intend to produce events to message bus.
+message Event {
```
StewartJingga, line 14: @ravisuhag do we still need this? it is not being used in `meteor` atm.

odpf/assets/facets/v1beta2/lineage.proto
GrayFlash: have removed the facets, and now the common has three proto files, `event`, `lineage`, and `owner`. I think it should be fine to have these three in different files for better separation of functionality.

odpf/assets/common/v1beta2/event.proto
ravisuhag: What about in kafka sink?

odpf/assets/facets/v1beta2/lineage.proto
ravisuhag: Reasoning i am thinking single file at the same level as asset is so that users in extractors does not have to import two packages. and it makes it little easier for them to use. thoughts?

StewartJingga: was thinking the same thing, but lineage is now also used in `Dashboard.Chart` model.
So that's why we decided to put it in common folder.

odpf/assets/common/v1beta2/event.proto
StewartJingga: We could but that is pretty specific use case (only when using kafka sink) and might not represent a metadata.
But then again, I'm not sure where else to put it 🤔 .

ravisuhag: We can avoid it from asset but then we need to think about how to handle this in kafka sink.

odpf/assets/facets/v1beta2/lineage.proto
ravisuhag: @StewartJingga How does that impact? I am thinking of a `common.proto` file at the asset level which holds these protos. So we still have these messages separate but no dedicated package is needed.

StewartJingga: @ravisuhag  Oh I thought you meant it to be inside asset file alongside `Asset` model.
Then yeah we can have all those in `common.proto`.

GrayFlash: commited the changes.

ravisuhag (approved):

PR #168, opened 2022-07-28
feat(compass): add new fields in asset model

scortier:
#### Added new fields on Asset to accomodate the new model from Meteor

1. `url: string`

2. `attributes: map[string]interface{}`

### DIFF
--- a/odpf/compass/v1beta1/service.proto
+++ b/odpf/compass/v1beta1/service.proto
@@ -911,27 +911,31 @@ message GetAssetByIDResponse {
 message UpsertPatchAssetRequest {
   message BaseAsset {
     string urn = 1;
-    string type = 2;
-    google.protobuf.StringValue name = 3 [
+    string url = 2;
+    string type = 3;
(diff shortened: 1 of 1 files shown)


StewartJingga:
@ravisuhag 

Meteor Asset model now has
1. Data `Table | Dashboard | Job`
2. Labels `map<string,string>`
3. Attributes `map of any`

While Compass in this PR will have
1. Data `map of any`
2. Labels `map<string,string>`
3. Attributes `map of any`

There could be a redundancy on Compass model if we add `Attributes` since it behaves the same as `Data`.

But without it, there could be conflicts if we merge `Meteor Data and Attributes` before mapping it to `Compass' Data`.

Do you think we should add this new `Attributes` field in `Compass`?

ravisuhag:
`Attributes` are mostly related to model metadata. So, they are better suited in `Table|Topic|Dashboard`. `Asset` on the other hand can still continue to maintain labels.

ravisuhag:
Closed in reference to a new PR. #174

PR #170, opened 2022-08-03
feat: add appeal config for cross individual appeal

singhvikash11:


bsushmith (approved):

mabdh:
@singhvikash11 @bsushmith  need more context here. to which issue in guardian does this PR for?

singhvikash11:
@mabdh this one is addressing this [issue](https://github.com/odpf/guardian/issues/212)

odpf/guardian/v1beta1/guardian.proto
```diff
@@ -541,6 +541,7 @@ message PolicyAppealConfig {
         string value = 2;
     }
     repeated DurationOptions duration_options = 1;
+    bool  allow_cross_individual_user = 2;
```
ravisuhag, line 544: @singhvikash11 can we think of a better name for this?

singhvikash11: @ravisuhag @ravisuhag these are names I can suggest:
allow_appeal_for_other
on_behalf_appeal

ravisuhag: How about on the lines of `recipient `,  `grantee`, `beneficiary ` ?

singhvikash11: recipient or grantee is fine with me.

@rahamat wdyt?

bsushmith: Any specific reason cons for `allow_on_behalf_appeal` or `allow_appeal_for_other`? The variable name itself would be very clear about what it means with minimal confusion. 🤔

singhvikash11: @ravisuhag @rahmatrhd can we agree on `allow_on_behalf_appeal`

ravisuhag: @bsushmith IMO, we need to find the right entity name to communicate core concepts.

rahmatrhd: I think `allow_creation_for_other_individual_user` is the most accurate & descriptive one even though it's too long.
1. `allow` word describing that it is an enable/disable option
2. mentions that it's for appeal creation
3. explains that the option is for **cross-individual** appeal creation only, because creating an appeal for a different account type (example: service account) is already supported and out of this field's scope.
`user` word can be replaced with `recipient`/`grantee`/etc. but the other words still need to be kept IMO

singhvikash11: @rahmatrhd renamed the variable to`allow_creation_for_another_user`

ravisuhag: @rahmatrhd @singhvikash11   I think it is too long. How about `allow_on_behalf`?

rahmatrhd: `allow_on_behalf_creation` doesn't explicitly describe that it's for cross-individual appeal but I'm good with it

singhvikash11: @rahmatrhd @ravisuhag renamed the variable to `allow_on_behalf`

PR #173, opened 2022-08-12
feat(optimus) : job explain api

Mryashbhardwaj:
needed for :  https://github.com/odpf/optimus/issues/492

### DIFF
--- a/odpf/optimus/core/v1beta1/job_spec.proto
+++ b/odpf/optimus/core/v1beta1/job_spec.proto
@@ -140,7 +140,6 @@ message JobExplainResponse {
 }
 
 message CreateJobSpecificationRequest {
-  reserved 3;
   string project_name = 1;
   string namespace_name = 2;
   JobSpecification specs = 3;
--- a/odpf/optimus/core/v1beta1/job_spec.proto
+++ b/odpf/optimus/core/v1beta1/job_spec.proto
@@ -148,7 +148,6 @@ message CreateJobSpecificationRequest {
 message CreateJobSpecificationResponse {
   bool success = 1;
   string message = 2;
-  string deployment_id = 3;
 }
 
 message GetJobSpecificationRequest {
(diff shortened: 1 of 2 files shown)


odpf/optimus/core/v1beta1/job_spec.proto
```diff
@@ -119,6 +127,53 @@ message AddJobSpecificationsResponse {
...
 }
 
+message JobInspectRequest {
+  string project_name = 1;
+  string namespace_name = 2;
+  string job_name = 3;
+  JobSpecification spec = 4;
+}
+
+message BasicInfoSection {
```
sravankorumilli, line 137: it would be better to keep all the dependent  messages with in the JobInspectResponse @sbchaos what do you say? Want to keep this very scoped as they won't be shared at all.

odpf/optimus/core/v1beta1/job_spec.proto
```diff
@@ -119,6 +127,53 @@ message AddJobSpecificationsResponse {
...
+  string destination =3;
+  repeated Log notice =4;
+}
+
+message OptimusDependency {
+  string name = 1;
+  string host = 2;
+  string project_name = 3;
+  string namespace_name = 4;
+  string task_name = 5;
```
sravankorumilli, line 149: what is this task_name? how is it different from name? Want to understand its relevance in dependency?

sravankorumilli: cc : @arinda-arif

sravankorumilli: Lets rename to JobDependency

sravankorumilli (changes requested):

odpf/optimus/core/v1beta1/job_spec.proto
Mryashbhardwaj: done

odpf/optimus/core/v1beta1/job_spec.proto
Mryashbhardwaj: task name is name of the scheduler task , vs name is job name

Mryashbhardwaj: there is already a `JobDependency` in proton, but that does not capture as much into

sravankorumilli (approved):

PR #176, opened 2022-08-18
feat: user upsert 

ishanarya0:
Allow update with id along with user upsert by email through `PUT /users/{id_or_email}`

Related Issue: https://github.com/odpf/shield/issues/162

### DIFF
--- a/odpf/shield/v1beta1/shield.proto
+++ b/odpf/shield/v1beta1/shield.proto
@@ -75,12 +75,12 @@ service ShieldService {
 
   rpc UpdateUser(UpdateUserRequest) returns (UpdateUserResponse) {
     option (google.api.http) = {
-      put: "/v1beta1/users/{id}",
+      put: "/v1beta1/users/{id_or_email}",
       body: "body";
     };
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
       tags: "User";
-      summary: "Update User by ID";
+      summary: "Update User by ID or Upsert by email";
     };
   }
 
@@ -661,7 +661,7 @@ message UpdateCurrentUserResponse {
 }
 
 message UpdateUserRequest {
-  string id = 1;
+  string id_or_email = 1;
   UserRequestBody body = 2;
 }
 


odpf/shield/v1beta1/shield.proto
```diff
@@ -75,12 +75,12 @@ service ShieldService {
 
   rpc UpdateUser(UpdateUserRequest) returns (UpdateUserResponse) {
     option (google.api.http) = {
-      put: "/v1beta1/users/{id}",
+      put: "/v1beta1/users/{id_or_email}",
```
krtkvrm, line 78: wdys about `identifier` or `queryIdentifier`? `id_or_email` imo seems not to be a good var name

krtkvrm: @spy16 @rohilsurana @mabdh @AkarshSatija  wdys ?

spy16: `identifier` is same as `id` right (in terms of information it provides).. 

i like to compare var-names based on useful information they provide (or reduction in ambiguity). based on that, i think `id_or_email` is good (`id_email` on the other hand still remains slightly ambiguous). since the API itself is getting bit ambiguous now (2 kinds of lookup keys), this definitely helps resolve that.

any particular reason why you have second thoughts about `id_or_email`?

ravisuhag: I think we can just keep it `id` itself. I see `id` as an identifier here.

spy16: why not explicit `id_or_email` though? why hide that information when showing it doesn't cost anything? (it's not really going to show up in our code or anything. but for anyone looking at the protobuf files, kinda acts as self-documentation)

mabdh: How would the go file be generated with this params? Would it become `request.GetIdOrEmail` ?

If yes, maybe in the future if we want to get rid the behaviour to upsert by Email, this will become confusing, or even if we want to add a behaviour `update user by user name`, somehow cannot reuse this (if we want to have a strict documentation) or maybe we need to create a new API.

ishanarya0: Yes, it would be `request.GetIdOrEmail`. Currently, the variable would be `id_or_email` in our code as well, and we'll check if it is an uuid or email, using shield'd uuid package etc. 

Later, if we want to remove the upsert behavior, I don't think it would be difficult. We just have to make changes in proto and the flow could use `UpdateByUsername` instead of `UpdateByEmail`.

spy16: if we have plans to remove this in the future (as a likely goal.. not just hypothetical this-might-happen assumption) then may be just having it as `id` and documenting somewhere would be okay i guess (because when that does happen, we will need proto changes to change this back into `id`)

ishanarya0: Yes, let's go ahead with just id then. 

Based on this discussion, I am closing this PR.

PR #178, opened 2022-08-22
feat(predator): add metricslog and resultlog

arinda-arif:


odpf/predator/v1beta1/metrics_log.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
+
+package odpf.predator.v1;
```
ravisuhag, line 3: @arinda-arif package name should be v1beta1

arinda-arif: thanks @ravisuhag, updated.

PR #180, opened 2022-08-29
feat: add create metadata key API

ishanarya0:
To support metadata key creation for this [shield issue](https://github.com/odpf/shield/issues/162)

### DIFF
--- a/odpf/shield/v1beta1/shield.proto
+++ b/odpf/shield/v1beta1/shield.proto
@@ -95,6 +95,18 @@ service ShieldService {
     };
   }
 
+  //Metadata Keys
+    rpc CreateMetadataKey(CreateMetadataKeyRequest) returns (rpc CreateMetadataKeyResponse) {
+    option (google.api.http) = {
+      post: "/v1beta1/metadatakey",
+      body: "body";
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      tags: "Metadata Key";
+      summary: "Create Metadata Key";
+    };
+  }
+
   // Group
   rpc ListGroups(ListGroupsRequest) returns (ListGroupsResponse) {
     option (google.api.http) = {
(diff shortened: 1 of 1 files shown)


odpf/shield/v1beta1/shield.proto
```diff
@@ -644,6 +656,24 @@ message CreateUserResponse {
   User user = 1;
 }
 
+message MetadataKeyRequestBody {
+  string key = 1 [(validate.rules).string.pattern = "^[A-Za-z0-9_-]+$"];
```
krtkvrm, line 660: How about key only having Upper Caps, like enum?

krtkvrm (changes requested):

odpf/shield/v1beta1/shield.proto
ishanarya0: This regex shall allow all uppercase keys. 
`ex: STOP, START etc would be valid`

I think it's the desired behavior(?)

krtkvrm: It will also allow - `Edit`, `slack-Id`

ishanarya0: @krtkvrm Can you please help in listing down the formats/examples of keys that should be accepted and not accepted?

ravisuhag: Should we just let me plain string? Any specific need to put validation on this apart from string?

krtkvrm: No, just regex that would be acceptable for key
Like `STOP_FIREHOSE` or `Stop-Firehose` or `StopFirehose`.. etc. I would prefer 1st one, open to discussions

ishanarya0: I wanted to keep a check on the use of some symbols that generally are not there in keys. Examples of such keys are` @type, stop(), 1==1`. By allowing pain text, users can use anything for the key. 

Shall we still allow plain text @ravisuhag ?

ishanarya0: @krtkvrm Yes, shall we allow all of these, including all small caps and snake case?  Ex `manager, manager_email`

ravisuhag: +1. I think we can just skip symbols and allow all other characters.

krtkvrm (approved):

PR #187, opened 2022-09-19
feat(compass): add CreateAssetProbe

StewartJingga:
Relates to https://github.com/odpf/compass/issues/170

### DIFF
--- a/odpf/compass/v1beta1/service.proto
+++ b/odpf/compass/v1beta1/service.proto
@@ -372,6 +372,20 @@ service CompassService {
     };
   }
 
+  rpc CreateAssetProbe(CreateAssetProbeRequest) returns (CreateAssetProbeResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/assets/{urn}/probes"
+    };
+
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "Create a new asset's probe";
+      description: "Add a new probe to an asset";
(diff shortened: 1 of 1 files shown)


odpf/compass/v1beta1/service.proto
```diff
@@ -1036,6 +1051,32 @@ message GetAssetByVersionResponse {
   Asset data = 1;
 }
 
+message CreateAssetProbeRequest {
+  message BaseProbe {
```
ravisuhag, line 1055: @StewartJingga any reason to calli it `BaseProbe` ?

StewartJingga: @ravisuhag  No major reason, mostly because I was following how we are doing it for other request payload as seen [here](https://github.com/odpf/proton/blob/main/odpf/compass/v1beta1/service.proto#L929).

Also the scope of this message only for `CreateAssetProbeRequest`, so if it needs to be changed I don't see any issue with it.

ravisuhag: I think in both the cases we should keep it without Base?  wdyt?

ravisuhag: Yes. lets do that.

StewartJingga: @ravisuhag updated `BaseProbe` to `Probe`

And here is the PR for `BaseAsset` to `Asset`
https://github.com/odpf/proton/pull/188

PR #193, opened 2022-09-27
feat: add metadata models to enable E2E lineage

sudo-suhas:
For enabling End-to-End lineage, add metadata models for the following:
- Machine Learning(ML) Feature
- ML Model
- Service
- Experiment
- Metric

### DIFF
--- a/odpf/assets/v1beta2/experiment.proto
+++ b/odpf/assets/v1beta2/experiment.proto
@@ -36,7 +36,7 @@ message Experiment {
   string entity = 1;
 
   // Optional: Percentage of the traffic that the experiment is enabled for.
-  float experiment_traffic = 2;
+  float traffic_percent = 2;
 
   // The variants of the experiment possibly including the control group.
   repeated Variant variants = 3;
--- a/odpf/assets/v1beta2/experiment.proto
+++ b/odpf/assets/v1beta2/experiment.proto
@@ -1,0 +1,67 @@
+syntax = "proto3";
+
+package odpf.assets.v1beta2;
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/odpf/proton/assets/v1beta2;assetsv1beta2";
+option java_outer_classname = "ExperimentProto";
+option java_package = "io.odpf.assets";
+
+// An experiment is the set of configurations and filters that allow for
+// systematically varying some independent variables to impact some other
+// dependent variables.
+message Experiment {
+  // Instance of configurations to be compared in the experiment.
+  message Variant {
+    // Name of the experiment variant.
+    string name = 1;
+
+    // Traffic percent enabled for the variant.
+    float traffic_percent = 2;
+
+    // Indicated whether the variant is the control for the experiment.
+    bool is_control = 3;
+
+    // List of properties the entity has.
+    google.protobuf.Struct attributes = 4;
+
+    // Whether the variant has been promoted to all users.
+    bool is_promoted = 5;
+  }
+
+  // Optional: Type of the entity being experimented over. ex: customer, session,
+  // device, driver etc.
+  string entity = 1;
+
+  // Optional: Percentage of the traffic that the experiment is enabled for.
+  float experiment_traffic = 2;
+
+  // The variants of the experiment possibly including the control group.
+  repeated Variant variants = 3;
+
+  // Optional: List of attributes the experiment has. This could include the
+  // following:
+  // - client_id[string]: The ID if the client running the experiment.
+  // - client_name[string]: The name of the client running the experiment.
+  // - primary_metric[string]: Used to determine a statistically significant
+  //   winning or losing variant.
+  // - guardrail_metric[string]: Business metric designed to indirectly measure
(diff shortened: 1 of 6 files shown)


odpf/assets/v1beta2/experiment.proto
```diff
@@ -0,0 +1,67 @@
...
+    // Whether the variant has been promoted to all users.
+    bool is_promoted = 5;
+  }
+
+  // Optional: Type of the entity being experimented over. ex: customer, session,
+  // device, driver etc.
+  string entity = 1;
+
+  // Optional: Percentage of the traffic that the experiment is enabled for.
+  float experiment_traffic = 2;
```
ravisuhag, line 39: @sudo-suhas is this duplicated?  line 22.

odpf/assets/v1beta2/ml_feature.proto
```diff
@@ -0,0 +1,86 @@
...
+option go_package = "github.com/odpf/proton/assets/v1beta2;assetsv1beta2";
+option java_outer_classname = "MLFeatureProto";
+option java_package = "io.odpf.assets";
+
+// MLFeature is a Machine Learning(ML) feature. In machine learning, a feature
+// is an individual measurable property, typically represented by a column,
+// that serves as an input for Machine Learning (ML) algorithms.
+message MLFeature {
+  // Feature value type. Used to define data types in Feature Tables.
+  enum DataType {
```
ravisuhag, line 16: @sudo-suhas do we need this enum? can this be just string?

sudo-suhas: The idea was to homogenize the types across different vendors who might represent the same underlying type slightly differently. Don't have a strong opinion though, can just use string if you think that would be better.

ravisuhag: @sudo-suhas I think as metadata owners we can't enforce it anyway. So better to just use what the source system provides and keep it flexible. So my opinion would be to remove it.

odpf/assets/v1beta2/experiment.proto
ravisuhag, line 39: @sudo-suhas shall we remove the experiment from here? since it is implied.

ravisuhag: Also lets make it consistent if we use percentage as keyword here and in variant traiffc.

odpf/assets/v1beta2/metric.proto
```diff
@@ -0,0 +1,42 @@
...
+  string namespace = 1;
+
+  // The field being used to calculate a metric.
+  string field_name = 2;
+
+  // Type of the evaluated metric. ex: count_distinct, average etc.
+  string measure_type = 3;
+
+  // Optional: The SQL representation of the query with filters and aggregations.
+  string sql = 4;
```
ravisuhag, line 25: @sudo-suhas lets call it query instead of SQL. Because the query interface could be anything and not just SQL.

odpf/assets/v1beta2/ml_feature.proto
```diff
@@ -0,0 +1,86 @@
...
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/odpf/proton/assets/v1beta2;assetsv1beta2";
+option java_outer_classname = "MLFeatureProto";
+option java_package = "io.odpf.assets";
+
+// MLFeature is a Machine Learning(ML) feature. In machine learning, a feature
+// is an individual measurable property, typically represented by a column,
+// that serves as an input for Machine Learning (ML) algorithms.
+message MLFeature {
```
ravisuhag, line 14: @sudo-suhas lets call it just feature.

odpf/assets/v1beta2/ml_feature.proto
```diff
@@ -0,0 +1,86 @@
...
+  string algorithm = 3;
+
+  // Optional: Name of the entity instance.
+  string entity_name = 4;
+}
+
+// MLFeatureTable is a Machine Learning(ML) feature table or view that
+// represents a logical group of time-series feature data as it is found in a
+// data source.
+message MLFeatureTable {
```
ravisuhag, line 54: @sudo-suhas lets call this just FeatureTable?

odpf/assets/v1beta2/service.proto
```diff
@@ -0,0 +1,23 @@
...
+option go_package = "github.com/odpf/proton/assets/v1beta2;assetsv1beta2";
+option java_outer_classname = "ServiceProto";
+option java_package = "io.odpf.assets";
+
+message Service {
+  // The service/application's ID
+  string application_id = 1;
+
+  // Optional: The version of the service.
+  string version = 2;
```
ravisuhag, line 16: @sudo-suhas let's add git url as well in service.

odpf/assets/v1beta2/ml_feature.proto
sudo-suhas: I feel 'feature' would not be clear enough as it is an overloaded word.

odpf/assets/v1beta2/service.proto
ravisuhag: yes. actually lets stick to that.

PR #195, opened 2022-10-03
feat(entropy): add log filtering labels to resource state

rohilsurana:


odpf/entropy/v1beta1/resource.proto
```diff
@@ -93,6 +101,7 @@ message ResourceState {
   Status status = 1;
   google.protobuf.Value output = 2;
   bytes module_data = 3;
+  LogOptions log = 4;
```
spy16, line 104: what do you think about the name `log_options` for the field?

rohilsurana: done

spy16 (approved):

PR #197, opened 2022-10-07
feat(guardian): import pre existing access

rahmatrhd:


odpf/guardian/v1beta1/guardian.proto
```diff
@@ -66,6 +66,13 @@ service GuardianService {
     };
   }
 
+  rpc ImportAccess(ImportAccessRequest) returns (ImportAccessResponse) {
+    option (google.api.http) = {
+      post: "/v1beta1/providers/{id}/import-access"
```
ravisuhag, line 71: @rahmatrhd How about just `import`?

rahmatrhd: @ravisuhag how about changing it to `POST /grants/import` then? In the request body, provider ID can be added in the body along with the other import criterias. Another reason is also the response from this API is []Grant

ravisuhag: This sounds better. Lets go ahead with it. Any thoughts? @mabdh @haveiss

rahmatrhd:
@ravisuhag @mabdh @haveiss applied the change here: https://github.com/odpf/guardian/pull/309

haveiss (approved):

odpf/guardian/v1beta1/guardian.proto
```diff
@@ -224,6 +217,13 @@ service GuardianService {
       body: "*"
     };
   }
+
+  rpc ImportAccess(ImportAccessRequest) returns (ImportAccessResponse) {
```
ravisuhag, line 221: @rahmatrhd how about naming it is as `ImportGrants` or `ImpotGrantsFromProvider`?

PR #200, opened 2022-10-11
feat(compass): add node attributes in CompassService.GetGraph RPC

sudo-suhas:
- The response now includes the node attributes with probes info for
  each source and target in the lineage edges.
- Deprecate type and service fields on LineageNode used in
  UpsertAssetRequest and UpsertPatchAssetRequest. These fields are no
  longer necessary for uniquely identifying an asset - the URN suffices.

### DIFF
--- a/odpf/compass/v1beta1/service.proto
+++ b/odpf/compass/v1beta1/service.proto
@@ -827,7 +827,19 @@ message GetGraphRequest {
 }
 
 message GetGraphResponse {
+  message ProbesInfo {
+    Probe latest = 1;
+  }
+
+  message NodeAttributes {
+    ProbesInfo probes = 1;
+  }
+
+  // Edges in the graph.
   repeated LineageEdge data = 1;
+  // Key is the asset URN. Node attributes, if present, will be returned for
+  // source and target nodes in the LineageEdge.
+  map<string, NodeAttributes> node_attrs = 2;
 }
 
 message GetAllTypesRequest {
@@ -1420,8 +1432,8 @@ message LineageNode {
   };
 
   string urn = 1;
-  string type = 2;
-  string service = 3;
+  string type = 2 [deprecated = true];
+  string service = 3 [deprecated = true];
 }
 
 message Tag {


odpf/compass/v1beta1/service.proto
```diff
@@ -244,6 +244,21 @@ service CompassService {
     };
   }
 
+  rpc GetLineageNodes(GetLineageNodesRequest) returns (GetLineageNodesResponse) {
+    option (google.api.http) = {
+      get: "/v1beta2/lineage/{urn=**}"
```
sudo-suhas, line 249: @ravisuhag do you see any concern with bumping the version for this endpoint from `v1beta1` to `v1beta2`? The existing endpoint on `/v1beta2/lineage/{urn=**}` returns the lineage graph edges.

cc: @StewartJingga

ravisuhag: @sudo-suhas yes, we should not increase version within the endpoint. If we ever have to create v1beta2. whole compass service would need to have v1beta2. So we should keep this v1beta1 only. Or create an entire v1beta2.

sudo-suhas: `/v1beta1/lineage/v2/{urn=**}` - is this okay?

ravisuhag: Any harm in deprecating the older one entirely and moving to this? Instead of keeping both?

StewartJingga: Mostly to prevent breaking changes, since the API is being used atm.

ravisuhag: I think we can document the migration strategy if we are deprecating the older version. Or should move to v1beta2. This approach breaks the convention.

sudo-suhas: I feel we should be able to accommodate newer version of _some_ specific endpoint with backward incompatible API without needing to change the version for all endpoints.

Furthermore, we should have a planned approach for the removal of any API where we mark it as deprecated for a release and remove it in a subsequent, but not necessarily the next, release. Directly removing in a single step will make it difficult for clients to migrate to the newer release because it blocks out-of-order usage of features and fixes.

sudo-suhas: Based on a discussion with @StewartJingga have made changes so that no breaking changes are necessary. The primary reason for doing so is that the current response is sympathetic to the graph storage format. Returning the lineage in the format proposed(https://github.com/odpf/compass/issues/170#issuecomment-1273116492) would need transformation at the application layer that would not be straightforward.

Marking this conversation as resolved, please do check the updated changeset.

ravisuhag: Looks good for now. But i think in long run we should think about considering the 170 discussion points.

PR #201, opened 2022-10-11
refactor(optimus): deprecated register instance service

deryrahman:


Mryashbhardwaj (approved):
looks good to me , lets add a card, to later remove the endpoint and the proto too

PR #212, opened 2022-11-02
feat: add custom question support

haveiss:
Related issue: https://github.com/odpf/guardian/issues/268

### DIFF
--- a/odpf/guardian/v1beta1/guardian.proto
+++ b/odpf/guardian/v1beta1/guardian.proto
@@ -610,6 +610,13 @@ message ProviderConfig {
   }
   repeated ResourceConfig resources = 6;
   repeated string allowed_account_types = 7;
+  message ProviderParameter {
+    string key = 1;
+    string label = 2;
+    bool required = 3;
+    string description = 4;
+  }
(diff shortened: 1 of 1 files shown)


odpf/guardian/v1beta1/guardian.proto
```diff
@@ -609,6 +609,13 @@ message ProviderConfig {
   }
   repeated ResourceConfig resources = 6;
   repeated string allowed_account_types = 7;
+  message ProviderParameter {
```
ravisuhag, line 612: @haveiss what will this be used for?

haveiss: Based on the issue and mas rahmat, it will be used whenever a new provider is registered via Guardian the custom question is defined for that during the creation itself.
Example use case for asking github username if the provider is github

CMIIW mas @rahmatrhd

ravisuhag: Yes, understood that. but this parameter is to collect questions.

ravisuhag (approved):

PR #220, opened 2022-11-10
feat: generate and publish the java protos jar 

AkbaraliShaikh:
Generate and Publish the Java Proto Jar 
Added the gradle build to generate and publish the proto jar


.github/workflows/build-java.yaml
```diff
@@ -0,0 +1,25 @@
+name: Build Java Protos
+
+on:
+  push:
+    branches:
+      - master
+      - main
+  pull_request:
```
rajathbk, line 8: Can remove these events if not relevant for this workflow..

AkbaraliShaikh: This is to just verify any build error on proto generating.

rajathbk (approved):

ravisuhag:
@rohilsurana Buf has not yet released the Java remote generation. Meanwhile any opinion on explicitly generating kar from proton ?

rohilsurana:
As of now we only have 2 options, 
 - to publish jar from this repo or
 - to clone proton to another repo and generate code there.

I see that this is only being used for compiling racoon protos, in that case the second approach would be preferred.

And if we decide to go with this approach we should publish the artifact with a different id than `proton` as it will be a bit misleading since it only includes racoon related schema/code.

java/build.gradle
```diff
@@ -0,0 +1,123 @@
...
+    mavenLocal()
+    maven {
+        url "https://plugins.gradle.org/m2/"
+    }
+}
+
+sourceSets {
+    main {
+        proto {
+            srcDir "${projectDir}/../odpf/raccoon/"
```
rohilsurana, line 35: if we plan to only publish racoon related schemas we should name the `groupId` and `artifactID` accordingly.

java/build.gradle
```diff
@@ -0,0 +1,123 @@
...
+plugins {
+    id 'java'
+    id "com.google.protobuf" version "0.9.1"
+    id 'maven-publish'
+    id 'signing'
+    id "io.github.gradle-nexus.publish-plugin" version "1.1.0"
+}
+
+group 'io.odpf'
+version '0.0.1'
```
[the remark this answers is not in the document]
rohilsurana: In case we want to manage it manually, lets just set it to the current release version.

java/build.gradle
AkbaraliShaikh: I tried generating the other protos but they were failing, 
so for the time being, I've only included the raccoon.

java/build.gradle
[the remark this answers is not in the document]
AkbaraliShaikh: updated here the version: https://github.com/odpf/proton/pull/228
please let me know if that is the right version.

java/build.gradle
ravisuhag: @AkbaraliShaikh +1 on this. I think we should publish all protos with proton. If it is just raccoon then we can use raccoon repo to generate code based on a proton commit. In the same way, we are doing this for go.

AkbaraliShaikh: I wanted to publish all the protos, not just the raccoon ones, but there was a problem with the dependent protos that were used by other protos in this repo.

In case of golang/buf we can simply provide the dependent list like below.
```
deps:
  - buf.build/googleapis/googleapis
  - buf.build/grpc-ecosystem/grpc-gateway
  - buf.build/envoyproxy/protoc-gen-validate
```

but in case of java (gradle), we might have to download all the dependent locally to generate.
It is not easy like the buf.

rohilsurana: We can use `buf` to take care of the java code generation instead of `protoc`. Once code is generated we can use `gradle` for packaging and publishing the jar.

ravisuhag: +1

chakravarthyvp: It'd be cleaner to have a multi-module project (either with gradle or buf way) where
the group can become `io.odpf.proton` (groupId) as the parent project and depends on each child's artefact project builds. The artifactId pushed is `raccoon|optimus|xxx`
This way in future, new children can built separately & bundle the way it needs to.

chakravarthyvp: If multi-module is time intensive to do now, we can adopt to move to building it in Racccon 

@AkbaraliShaikh

ravisuhag: +1 I think we are thinking of just compiling raccoon now, we use generate code within raccoon java client using buf itself and package into raccoon client jar. and convert this pipeline to entire proton pipeline. Later on if needed we can setup per project workflow.

PR #223, opened 2022-11-15
feat!: support multiple model versions in a single asset

sudo-suhas:
BREAKING CHANGES: Structure of Model proto is changed in a backward incompatible manner.


sudo-suhas:
Thanks, @ravisuhag!

PR #229, opened 2022-11-22
feat(guardian): add provider activity

rahmatrhd:


haveiss (commented):
please help to rebase from main branch

odpf/guardian/v1beta1/guardian.proto
```diff
@@ -291,6 +310,39 @@ message ListRolesResponse {
   repeated Role roles = 1;
 }
 
+message ImportActivitiesRequest {
+  string provider_id = 1;
+  repeated string account_ids = 2;
+  repeated string resource_ids = 3;
+  google.protobuf.Timestamp timestamp_gte = 4;
```
ravisuhag, line 317: @rahmatrhd So user would need to provide at least one of the timestamp or both?

rahmatrhd: both are optional. And since we are going to import activities using job, this will be used to filter the activities according to the job interval.

ravisuhag: Ok. That makes sense.

haveiss (approved):

PR #234, opened 2023-01-07
feat(siren): add create alert with namespace api

mabdh:
Related with https://github.com/odpf/siren/issues/163

### DIFF
--- a/odpf/siren/v1beta1/siren.proto
+++ b/odpf/siren/v1beta1/siren.proto
@@ -287,6 +287,19 @@ service SirenService {
     };
   }
 
+  rpc CreateAlertsWithNamespace(CreateAlertsRequest) returns (CreateAlertsResponse) {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "create alerts with namespace";
+      tags: "Alert";
+    };
+
+    option (google.api.http) = {
+      post: "/v1beta1/alerts/{provider_type}/{provider_id}/{namespace_id}"
+      body: "body"
+    };
+  }
+
+
   rpc ListRules(ListRulesRequest) returns (ListRulesResponse) {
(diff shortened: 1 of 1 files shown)


odpf/siren/v1beta1/siren.proto
```diff
@@ -287,6 +287,19 @@ service SirenService {
...
   }
 
+  rpc CreateAlertsWithNamespace(CreateAlertsWithNamespaceRequest) returns (CreateAlertsWithNamespaceResponse) {
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      summary: "create alerts with namespace";
+      tags: "Alert";
+    };
+
+    option (google.api.http) = {
+      post: "/v1beta1/alerts/{provider_type}/{provider_id}/{namespace_id}"
```
ravisuhag, line 297: @mabdh what blocks us from having `POST` on alerts with path `alerts/` and the rest of the parameters can be accepted with the body. Do we have a completely different structure of payload for each provider?

mabdh: @ravisuhag problem is we need to add this api in Cortex `webhook_config` url to send alerts and the body can't be configured and does not have namespace information, what we can do is just assign different url for different tenant.

mabdh: as an alternative, maybe we could just use `/v1beta1/alerts/{namespace_id}` ? considering provider type and provider id would be unique to each namespace?

mabdh: > Do we have a completely different structure of payload for each provider?
I think we need to expect the payload to be different for each provider and design the API to be extendable. We did that so each provider has a function to parse the consumed request body.

ravisuhag: Make sense. I think lets go ahead with this for now. And we can revisit this later.

StewartJingga (approved):

PR #240, opened 2023-01-18
refactor: removed APIs that are replaced by relation APIs

ishanarya0:
The following API routes shall go away, as shield now have relation APIs.

RESOURCE
```
AddTeamToResource
AddOwnerToResource
```
GROUP
```
AddUsers 
RemoveUser 
AddAdmins 
RemoveAdmin 
ListGroupUsers 
ListGroupAdmins 
```
ORGANIZATION
```
AddAdmins
RemoveAdmin
```
PROJECT
```
AddAdmins
RemoveAdmin
```

### DIFF
--- a/odpf/shield/v1beta1/shield.proto
+++ b/odpf/shield/v1beta1/shield.proto
@@ -148,68 +148,6 @@ service ShieldService {
     };
   }
 
-  rpc ListGroupUsers(ListGroupUsersRequest) returns (ListGroupUsersResponse) {
-    option (google.api.http) = {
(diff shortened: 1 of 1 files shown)


krtkvrm (approved):

kushsharma:
Wanted to understand the thought process behind keeping relations API and removing all these. Was the idea behind making it more flexible? If yes, what if we keep these specific APIs along with `relation` API? Because now the burden of converting these basic domain-level requests from generic relations comes to clients.

We could behind the scenes convert these basic use cases to relations?

PR #242, opened 2023-02-11
fix(shield): add validation for user slug

kushsharma:
- Making the body of user create/update mandatory and adding additional slug validation rules.
- Relaxing name constraint to contain space.


krtkvrm:
we are now working on coming up with `v1beta2`, thus will drop developing `v1beta1` unless fix is required

PR #243, opened 2023-02-13
feat(compass): support pagination in SearchAssets

bsushmith:
* Add pagination support in SearchAssets API. `from` would be the offset and `size` would be the page size. The changes are backward compatible

### DIFF
--- a/odpf/compass/v1beta1/service.proto
+++ b/odpf/compass/v1beta1/service.proto
@@ -801,13 +801,13 @@ message SearchAssetsRequest {
     ignore_empty: true
   }];
 
-  uint32 from = 7 [
+  uint32 offset = 7 [
     (validate.rules).uint32 = {
       gte: 0,
       ignore_empty: true
     },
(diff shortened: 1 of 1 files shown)


ravisuhag:
@bsushmith Let's discuss this please. It seems search is moving more and more towards list API. We should have clear responsibilities for list vs search.

bsushmith:
Irrespective of the search vs list API's discussion - the current search returns back a list of results with no flexibility by caller on how many results they would want to show or how they would want to display. Pagination/cursor should be something that should have been already present. This only fixes that by adding a backward compatible change.

sudo-suhas (approved):
LGTM

odpf/compass/v1beta1/service.proto
```diff
@@ -800,6 +800,16 @@ message SearchAssetsRequest {
     unique: true,
     ignore_empty: true
   }];
+
+  uint32 from = 7 [
```
sudo-suhas, line 804: Can we rename this to `offset` to keep it consistent with `GetAllAssetsRequest1`?

PR #246, opened 2023-02-27
feat: support for namespace in compass

kushsharma:
Compass is a single-tenant application, to make it multi-tenant just like most `odpf` applications few breaking changes are introduced.


rohilsurana:
Any specific reason to go with `namespace` as the variable for tenants. Can we instead use `project` as we want to isolate on basis of project?

kushsharma:
I am okay with `project` as GCP uses the same term, as long as we standardize the terminology across all products. Although I have seen `namespace` term being used for multi-tenancy as well quite a lot.

ravisuhag:
Agree that projects can be very specific. I think namespace or workspace are better choice.

PR #260, opened 2023-04-28
feat(shield): split shield apis separately for admin

kushsharma:


ravisuhag:
@kushsharma Few thoughts I ma having about keeping `/org` apis in admin. Other possible approach. 

- `/org` can be part of general apis as well. Shield organisation can have a privacy setting about making it public or not. and list api can list all public organisations. An example of this could be seen in Github org list apis. https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations
- Similar thing can be followed for `/users`. 
- We can also think about completely removing `/projects` since all projects actually belong to a organisations. 

Thoughts?

kushsharma:
> * `/org` can be part of general apis as well. Shield organisation can have a privacy setting about making it public or not. and list api can list all public organisations. An example of this could be seen in Github org list apis. https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations
> * Similar thing can be followed for `/users`.

Sure, looks alright to me. Do you mean this flag will be per organization and per user?

odpf/shield/v1beta1/admin.proto
```diff
@@ -0,0 +1,150 @@
...
+  };
+  schemes: HTTP;
+  schemes: HTTPS;
+  base_path: "/v1beta1";
+};
+service AdminService {
+  // Users
+  rpc ListAllUsers(ListAllUsersRequest) returns (ListAllUsersResponse) {
+    option (google.api.http) = {
+      get: "/users"
```
ravisuhag, line 28: @kushsharma we should add `/admin` prefix in apis here itself.

odpf/shield/v1beta1/shield.proto
```diff
@@ -86,7 +88,7 @@ service ShieldService {
 
   rpc UpdateCurrentUser(UpdateCurrentUserRequest) returns (UpdateCurrentUserResponse) {
     option (google.api.http) = {
-      put: "/v1beta1/users/self",
+      put: "/users/self",
```
ravisuhag, line 91: @kushsharma this api path can be just `/user`

ravisuhag:
> > * `/org` can be part of general apis as well. Shield organisation can have a privacy setting about making it public or not. and list api can list all public organisations. An example of this could be seen in Github org list apis. https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations
> > * Similar thing can be followed for `/users`.
> 
> Sure, looks alright to me. Do you mean this flag will be per organization and per user? We will have a `ListPublicOrganizations` instead of `ListOrganizations` in non-admin APIs?

We will have `ListOrganizations` in no-admin apis. and `ListAllOrganizations` in admin api.

odpf/shield/v1beta1/admin.proto
```diff
@@ -0,0 +1,148 @@
...
+    title: "Shield Administration API",
+    version: "0.2.0";
+  };
+  schemes: HTTP;
+};
+service AdminService {
+  // Users
+  rpc ListAllUsers(ListAllUsersRequest) returns (ListAllUsersResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/users"
```
ravisuhag, line 26: @kushsharma shouldn't admin also come in this prefix path? Also any reason to move `v1beta1` back to path?

ravisuhag (approved):

PR #264, opened 2023-05-01
feat(shield): add validation rules for org/project filters

kushsharma:


kushsharma:
Merging with breaking changes as they are unused fields.

PR #267, opened 2023-05-10
feat(shield): shield support for user-defined roles

kushsharma:
WIP

- https://github.com/odpf/shield/pull/240

### DIFF
--- a/odpf/shield/v1beta1/shield.proto
+++ b/odpf/shield/v1beta1/shield.proto
@@ -63,6 +63,16 @@ service ShieldService {
     };
   }
 
+  rpc ListCurrentUserGroups(ListCurrentUserGroupsRequest) returns (ListCurrentUserGroupsResponse) {
+    option (google.api.http) = {
+      get: "/v1beta1/users/self/groups",
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      tags: "User";
+      summary: "List groups of a User";
+    };
+  }
+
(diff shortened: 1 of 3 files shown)


odpf/shield/v1beta1/shield.proto
```diff
@@ -225,7 +245,7 @@ service ShieldService {
   // Roles
   rpc ListRoles(ListRolesRequest) returns (ListRolesResponse) {
     option (google.api.http) = {
-      get: "/v1beta1/roles",
+      get: "/v1beta1/organizations/{org_id}/roles",
```
ravisuhag, line 248: @kushsharma roles and permissions should be at root right?

kushsharma: Permissions are same system-wide but roles can be custom under each org I think?

kushsharma: I guess I can create a just `/roles/` endpoint as well to list predefined roles only.

ravisuhag: Yes, let's do that.  And `/org/id/role` would return predefined roles as well as custom for each org as well? 

Then I think we should only introduce `/org/id/roles` when we provide custom role creation api for an org.

kushsharma: I think `/org/id/role` should return only custom roles as that's what the API asks for. We will try our predefined roles are already very granular that this API is minimally used.

Also, the current CreateRole/UpdateRole API is actually CreateOrganizationRole API, let me rename it name as such, its uri already reflects it though.

PR #287, opened 2023-06-26
docs: update api summary and security schemes

Chief-Rishab:
- required for https://github.com/raystack/shield/pull/270


raystack/shield/v1beta1/admin.proto
```diff
@@ -195,7 +211,7 @@ service AdminService {
     };
     option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
       tags: "Resource";
-      summary: "Get all resources";
+      summary: "List Shield Resources";
```
ravisuhag, line 214: @Chief-Rishab lets not upper case each word. And any reason to include shield in description? I think we should avoid it.

kushsharma (approved):

PR #295, opened 2023-08-02
feat: add org domain whitelisting

Chief-Rishab:


raystack/frontier/v1beta1/frontier.proto
```diff
@@ -706,6 +706,75 @@ service FrontierService {
...
+    };
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      tags: "Organization";
+      summary: "Verify org domain";
+      description: "Verify a domain for an organization with a verification token generated by Frontier GenerateDomainVerificationToken API. The token must be added to your domain's DNS provider as a TXT record before it can be verified. This API returns the state of the domain (pending/verified) after verification.";
+    };
+  }
+
+  rpc ListOrganizationsByDomain(ListOrganizationsByDomainRequest) returns (ListOrganizationsByDomainResponse) {
+    option (google.api.http) = {get: "/v1beta1/organizations/domains/{name}"};
```
kushsharma, line 761: I wonder when we will use it directly, what if instead of putting it with domain name, we put it under user?
Like `/users/{id}/organizations` or `/users/self/organizations` and the response could be
```
message GetOrganizationsByUserResponse {
  repeated Organization organizations = 1;
  repeated Organization joinable_via_domain = 2;
}
```

kushsharma (approved):
LGTM

PR #316, opened 2023-10-24
feat: update guardian api docs

Chief-Rishab:


raystack/guardian/v1beta1/guardian.proto
```diff
@@ -430,10 +744,21 @@ message ListAppealsRequest {
   repeated string resource_urns = 7;
   repeated string order_by = 8;
   string created_by = 9;
+  uint32 size = 10 [(validate.rules).uint32 = {
```
kushsharma, line 747: let's keep them `int` only. There is no real advantage of making them `uint` and it adds unnecessary discrepancy between other ints.

Chief-Rishab: We are using the same in other cases as well. Should I replace all the 29 occurrences or only the ones I have added in this PR for filters?

kushsharma: Oh I did not notice we use it in the rest of the places of guardian. Then I think it's fine.

kushsharma (approved):

PR #369, opened 2024-08-28
add pagination to list all invoice API

AmanGIT07:
Adds pagination to list all invoice API. `page_num` and `page_size` params can be used to modulate the pages. `count` returns the total number of items (not the count of items in response) so that it can be consumed for different logics if needed.


anujk14 (approved):
lgtm

PR #408, opened 2025-05-05
feat: add v3 annotation

whoAbhishekSah:


rohilsurana:
not required anymore, as we are migrating to connectrpc

PR #421, opened 2025-08-19
Feat/session management

paanSinghCoder:


raystack/frontier/v1beta1/admin.proto
```diff
@@ -1699,3 +1705,18 @@ message GetCurrentAdminUserResponse {
   User user = 1;
   ServiceUser service_user = 2;
 }
+
+message ListUserSessionsRequest {
+  string user_id = 1;
+}
+
+message ListUserSessionsResponse {
+  repeated Session user_sessions = 1;
```
rsbh, line 1714: Minor nitpick, we are using `sessions` and `user_sessions` for the key. Can we use the same key name to keep consistency

paanSinghCoder: Yes, just`sessions` make more sense.

PR #424, opened 2025-09-10
fix(session): add user id validation for user id in ListUserSession

paanSinghCoder:


raystack/frontier/v1beta1/admin.proto
```diff
@@ -1707,7 +1707,9 @@ message GetCurrentAdminUserResponse {
 }
 
 message ListUserSessionsRequest {
-  string user_id = 1;
+  string user_id = 1 [
+    (validate.rules).string.min_len = 1
```
AmanGIT07, line 1711: If we need this to be required field and uuid, we can use the validation rule: 
[
    (google.api.field_behavior) = REQUIRED,
    (validate.rules).string.uuid = true];
]

(Since user exists in frontier and it uses uuid, we can add the uuid validation)

AmanGIT07: min_len will also work btw

PR #483, opened 2026-04-23
refactor(frontier): remove deprecated fields from list member request messages

rohilsurana:
## Summary
- Remove `permission_filter` field from `ListOrganizationUsersRequest` and `ListProjectUsersRequest`, reserve field ids
- Remove `with_roles` field from `ListOrganizationUsersRequest`, `ListProjectUsersRequest`, `ListProjectServiceUsersRequest`, `ListProjectGroupsRequest`, and `ListGroupUsersRequest`, reserve field ids

## Test plan
- [ ] Verify buf lint passes
- [ ] Verify downstream services compile against updated proto

### DIFF
--- a/raystack/frontier/v1beta1/frontier.proto
+++ b/raystack/frontier/v1beta1/frontier.proto
@@ -1603,9 +1603,8 @@ message ListOrganizationAdminsResponse {
 
 message ListOrganizationUsersRequest {
   string id = 1 [(buf.validate.field).string.min_len = 3];
-  reserved 2;
+  reserved 2, 3;
 
-  bool with_roles = 3;
   repeated string role_filters = 4;
 }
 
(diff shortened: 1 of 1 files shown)


raystack/frontier/v1beta1/frontier.proto
```diff
@@ -1602,9 +1602,8 @@ message ListOrganizationAdminsResponse {
 
 message ListOrganizationUsersRequest {
   string id = 1 [(buf.validate.field).string.min_len = 3];
-  string permission_filter = 2 [deprecated = true];
+  reserved 2, 3;
 
-  bool with_roles = 3;
   repeated string role_filters = 4;
```
whoAbhishekSah: Should we make it a UUID? As in, the frontend must send the role IDs, not the names?

whoAbhishekSah (approved):

raystack/frontier/v1beta1/frontier.proto
rohilsurana: we can make this change, looks like it is not being used any where as of now

rohilsurana: done

whoAbhishekSah (approved):

PR #498, opened 2026-08-12
fix(frontier): allow omitted plan state

RushikeshBhavsar3605:
## Problem

`PlanRequestBody.state` is currently validated as either `active` or `inactive`.
Because the zero value (`""`) is not ignored, `CreatePlan` requests that omit
`state` are rejected before reaching Frontier's create logic.

Frontier already defaults an omitted plan state to `active` when creating a
plan.

## Change

Update `PlanRequestBody.state` validation to ignore the zero value while
continuing to validate non-empty values against `active` and `inactive`.

## Context

Related to raystack/frontier#1853

### DIFF
--- a/raystack/frontier/v1beta1/admin.proto
+++ b/raystack/frontier/v1beta1/admin.proto
@@ -457,11 +457,14 @@ message PlanRequestBody {
   int64 trial_days = 7 [(buf.validate.field).int64 = {gte: 0}];
 
   // known states are "active" and "inactive"
-  string state = 8 [(buf.validate.field).string = {
-    in: [
-      "active",
-      "inactive"
-    ]
+  string state = 8 [(buf.validate.field) = {
+    ignore: IGNORE_IF_ZERO_VALUE
+    string: {
+      in: [
+        "active",
+        "inactive"
+      ]
+    }
   }];
 
   google.protobuf.Struct metadata = 20;


whoAbhishekSah:
Thanks for the PR and the clear write-up.

I looked into this, and the strict validation is deliberate. There was a discussion on #496 about this exact field: https://github.com/raystack/proton/pull/496#discussion_r3711650584

The short version: `CreatePlan` and `UpdatePlan` requests should say the state explicitly, instead of relying on a hidden default. These plan write APIs are part of a declarative flow, where the request describes the full desired state of the plan. The omitted-state case is handled on the client side: Frontier's reconcile tool fills in `active` when the plans file omits state, before it calls the API:

- https://github.com/raystack/frontier/blob/1e7ad49dda65008c180eb11f04b14ba0b3329722/internal/reconcile/billingplan.go#L235-L244
- https://github.com/raystack/frontier/blob/1e7ad49dda65008c180eb11f04b14ba0b3329722/internal/reconcile/billingplan_reconciler.go#L249

So the API can stay strict without making plan files verbose.

Because of that, I think we should keep the validation as it is. Sorry that this wasn't easy to discover; the decision only lives in that review thread.

If you'd like to contribute, part 2 of raystack/frontier#1853 (empty product description failing at Stripe) is still open, if you want to pick that up.

RushikeshBhavsar3605:
Thanks for the clarification. I understand that the strict validation is intentional and that the reconcile flow handles the omitted state before calling the API. I’ll leave this change aside and look into part 2 of Frontier #1853.

PR #500, opened 2026-08-20
chore!: remove deprecated name and namespace fields from frontier permission protos

whoAbhishekSah:
## Summary
Removes the deprecated `name` and `namespace` fields from the frontier `Permission` message (`models.proto`) and `PermissionRequestBody` (`admin.proto`). The `key` field (`service.resource.verb`) is the sole way to express a permission's identity in both requests and responses. Final proto step of raystack/frontier#1782.

## Why this is safe
- frontier stopped reading the request fields (raystack/frontier#1885) and no longer needs the response fields since `key` round-trips losslessly (raystack/frontier#1887).
- Field numbers and names are reserved, so they cannot be reused and `buf breaking` (WIRE) passes.
- On the wire, proto3 ignores unknown fields: an old client that still sends `name`/`namespace` gets the same behavior it gets today (the server no longer reads them). Clients that read the removed response fields must move to `key`; frontier will call this out in its release notes.

## Changes
- `raystack/frontier/v1beta1/models.proto`: drop `Permission.name` (2) and `Permission.namespace` (6); reserve the numbers and names; document `key`.
- `raystack/frontier/v1beta1/admin.proto`: drop `PermissionRequestBody.name` (1) and `.namespace` (2); reserve the numbers and names.

## Test Plan
- [x] `buf lint` passes
- [x] `buf breaking --against main` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/raystack/frontier/v1beta1/admin.proto
+++ b/raystack/frontier/v1beta1/admin.proto
@@ -293,14 +293,13 @@ message DeleteRoleRequest {
 message DeleteRoleResponse {}
 
 message PermissionRequestBody {
-  string name = 1 [deprecated = true];
-  // namespace should be in service/resource format
-  string namespace = 2 [deprecated = true];
+  reserved 1, 2;
+  reserved "name", "namespace";
+
   google.protobuf.Struct metadata = 3;
   string title = 4;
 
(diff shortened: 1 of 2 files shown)


rohilsurana:
Check `CheckResourcePermissionRequest` message in `frontier.proto` files as well.

rohilsurana (approved):

